PR #14, opened 2014-12-20
Add subprojects; first pass at adding laws and tests.

non:
This commit creates three sub-projects:

 - core: builds algebra.jar; contains type classes (and a few types)
 - std:  builds algebra-std.jar; contains sample instances
 - laws: builds algebra-laws.jar; contains laws and conformance tests

It also ports over the basic laws from Spire for:

 - signed
 - eq/order
 - groups
 - rings

Finally, it adds a very simple standard algebra instance for Int, and
runs the relevant laws tests on it.


non:
Most recent pushes merge the support for commutativity to this branch, and update the laws accordingly.

The nice thing is that the previous PR couldn't have tests -- but in this branch we were able to confirm that all the instances we had defined were in fact commutative! :fireworks:

tixxit:
:+1:

non:
What do you think @johnynek? Anything else you'd like to see done on this PR?

laws/src/main/scala/algebra/laws/package.scala
```diff
@@ -0,0 +1,47 @@
...
+    new Predicate[A] {
+      def apply(a: A) = ev.neqv(a, A.zero)
+    }
+
+  lazy val proved = Prop(Result(status = Proof))
+
+  lazy val falsified = Prop(Result(status = False))
+
+  object Ops {
+    def run[A](sym: String)(lhs: A, rhs: A)(f: (A, A) => Boolean): Prop =
```
larsrh: Why is this wrapped in an object? Couldn't this be just a regular function? Sorry I'm late to the party :-)

PR #19, opened 2014-12-29
Added Band, sorted the README table

johnynek:
I sorted the table by number of properties and then by column order of the first property satisfied.


non:
This looks good to me. Do you think we should add laws for `Band[A]`?

johnynek:
I do agree we need some laws. I'll update the PR and add them.

johnynek:
@non what do you think now?

tixxit:
Is it worth adding a method like `isIdempotent: Boolean` to `Semigroup` to mirror the `isCommutative` one? I'm not sure if this is something someone would want to check/optimize for.

non:
Looks fine to me. :+1: 

If @johnynek thinks having `def isIdempotent: Boolean` on `Semigroup[A]` would be useful, then I'm definitely fine with adding that as well. I'm not sure how many algorithms support optional idempotence though.

I definitely don't think it's worth adding idempotence-methods to the traits in `Additive.scala` and `Multiplicative.scala` though.

johnynek:
I'm not crazy about the isIdempotent.

What does that gain us over:

```scala
def isIdempotent(s: Semigroup[_]): Boolean = s.isInstanceOf[Band[_]]
```
and we don't clutter our traits with accessor methods.

What if we added those checks to `object Semigroup` so we could do:

```scala
object Semigroup {
  def isCommutative[A](implicit sg: Semigroup[A]): Boolean = sg.isInstanceOf[CommutativeSemigroup[A]]
  def isIdempotent[A](implicit sg: Semigroup[A]): Boolean = sg.isInstanceOf[Band[A]]
}
```

tixxit:
Yeah, I like that. :+1: on the PR

johnynek:
Looks like 2 :+1: on this. I'm clicking merge. Thanks for the reviews.

PR #27, opened 2015-01-06
Fix algebraic properties link

ceedubs:
Also use only header names instead of full URLs in links to make it a
bit less brittle (helps with multiple branches, forks, etc).


non:
:+1: from me

non:
(We're using a system where at least two people have to sign off on a PR to get it merged. So this PR has 1/2 now.)

PR #28, opened 2015-01-06
Fix broken link in README

igstan:


igstan:
Ah, yes. Sorry I missed that.

PR #29, opened 2015-01-07
This commit adds a rule that tests serializability.

### LINKED ISSUE #13: Ensure all type classes are serializable.
I think for use with Spark and possibly other frame works, it's very convenient if the instances can be serialized.

What do you all think? Does this make sense?

non:
It also adds the Serializable trait to all the
algebraic type classes.

Fixes #13.

### DIFF
--- a/laws/src/main/scala/algebra/laws/LatticeLaws.scala
+++ b/laws/src/main/scala/algebra/laws/LatticeLaws.scala
@@ -23,6 +23,7 @@ trait LatticeLaws[A] extends Laws {
   def joinSemilattice(implicit A: JoinSemilattice[A]) = new LatticeProperties(
     name = "joinSemilattice",
     parents = Nil,
+    Rules.serializable(A),
     Rules.associativity(A.join),
     Rules.commutative(A.join),
     Rules.idempotence(A.join)
@@ -31,6 +32,7 @@ trait LatticeLaws[A] extends Laws {
   def meetSemilattice(implicit A: MeetSemilattice[A]) = new LatticeProperties(
     name = "meetSemilattice",
     parents = Nil,
+    Rules.serializable(A),
     Rules.associativity(A.meet),
     Rules.commutative(A.meet),
     Rules.idempotence(A.meet)
--- a/laws/src/main/scala/algebra/laws/Rules.scala
+++ b/laws/src/main/scala/algebra/laws/Rules.scala
@@ -11,7 +11,24 @@ object Rules {
 
   def serializable[M](m: M): (String, Prop) =
     "serializable" -> Prop { _ =>
-      Result(status = if (m.isInstanceOf[Serializable]) Proof else False)
+      import java.io._
+      val baos = new ByteArrayOutputStream()
+      val oos = new ObjectOutputStream(baos)
+      var ois: ObjectInputStream = null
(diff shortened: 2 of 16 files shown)


laws/src/main/scala/algebra/laws/LatticeLaws.scala
```diff
@@ -23,6 +23,7 @@ trait LatticeLaws[A] extends Laws {
   def joinSemilattice(implicit A: JoinSemilattice[A]) = new LatticeProperties(
     name = "joinSemilattice",
     parents = Nil,
+    Rules.serializable(A),
```
tixxit: `JoinLattice` (and `MeetLattice`) both don't extend `Serializable`, while the others do. Should those also extend `Serializable`?

non: Yes they should! It's troubling that this test was passing. I'll take a closer look.

non:
Fixed!

laws/src/main/scala/algebra/laws/Rules.scala
```diff
@@ -3,11 +3,17 @@ package laws
...
 import org.scalacheck.Prop._
+import Prop.{False, Proof, Result}
 
 import algebra.std.boolean._
 
 object Rules {
 
+  def serializable[M](m: M): (String, Prop) =
+    "serializable" -> Prop { _ =>
+      Result(status = if (m.isInstanceOf[Serializable]) Proof else False)
```
johnynek: This really can't fail now since we are extending serializable in the base case. A stronger test would be to serialize and deserialize and not see an exception. That's a stronger constraint, but without doing that the serializable marker is not very useful.

std/src/main/scala/algebra/std/int.scala
```diff
@@ -8,6 +9,11 @@ package object int extends IntInstances
 
 trait IntInstances {
   implicit val intAlgebra = new IntAlgebra
+
+  val IntMinMaxLattice = new Lattice[Int] {
```
johnynek: stylewise, this looks weird. While I personally often do this, we seem to have a named class style so far. Curious why this one is anonymous.

laws/src/main/scala/algebra/laws/Rules.scala
non: That's a good point. I'll try to rig up something with `DataOutputStream` and `DataInputStream` to test actual serialization. Does that seem reasonable?

std/src/main/scala/algebra/std/int.scala
non: No real reason, I just threw this together for testing purposes. I'll probably generalize it to anything that has an order, TBH.

laws/src/main/scala/algebra/laws/Rules.scala
non: This should be fixed (and the more rigorous check did catch an issue with `std.Rat`).

johnynek:
Can we merge #33 into this (after it is in develop) and assuming it is green, I say merge.

non:
OK, #33 was merged into master, and then into this PR. Tests running now.

johnynek:
Sweet! Nice and green. +1

PR #30, opened 2015-01-09
Add miniboxed versions of artifacts.

tixxit:
This does a source based version (use same sources, but with a minor `algebra.spec` compat layer that changes between the 2). Basically, we use 2 type aliases, `@sp` and `@mb` for specialization and miniboxing respectively. Depending on whether we are using the standard versions or the `-miniboxed` versions, we'll either stub out the `@mb` or the `@sp` annotation. So, if miniboxing is enabled, we don't specialize. If specialization is enabled, we don't minibox.

The miniboxed versions don't currently compile. @VladUreche this is failing because the typeclasses extend `Any`. Is this a known limitation or can it be fixed?

I've also taken the opportunity to clean up the build a bit.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -27,47 +27,38 @@ scalacOptions ++= (
   Nil
 )
 
+// extra repositories (for miniboxing)
+
 resolvers in ThisBuild += Resolver.sonatypeRepo("snapshots")
 
-// publishing/release details follow
+// publish pre-amble
 
-publishMavenStyle := true
-publishArtifact in Test := false
-pomIncludeRepository := { _ => false }
+Publish.preamble
 
-publishTo <<= version { (v: String) =>
-  val nexus = "https://oss.sonatype.org/"
-  if (v.trim.endsWith("SNAPSHOT"))
-    Some("snapshots" at nexus + "content/repositories/snapshots")
-  else
-    Some("releases"  at nexus + "service/local/staging/deploy/maven2")
-}
+// projects
 
-pomExtra := (
-<scm>
-  <url>git@github.com:non/algebra.git</url>
-  <connection>scm:git:git@github.com:non/algebra.git</connection>
-</scm>
-<developers>
-  <developer>
-    <id>johnynek</id>
-    <name>P. Oscar Boykin</name>
-    <url>https://github.com/johnynek/</url>
-  </developer>
-  <developer>
-    <id>avibryant</id>
-    <name>Avi Bryant</name>
-    <url>https://github.com/avibryant/</url>
-  </developer>
-  <developer>
-    <id>non</id>
-    <name>Erik Osheim</name>
-    <url>http://github.com/non/</url>
-  </developer>
-  <developer>
-    <id>tixxit</id>
-    <name>Tom Switzer</name>
-    <url>http://github.com/tixxit/</url>
-  </developer>
-</developers>
-)
+lazy val aggregate = project.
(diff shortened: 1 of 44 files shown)


tixxit:
@VladUreche specifically, I'm getting errors like this:

```
[error] /Users/tom/Code/algebra/core/src/main/scala/algebra/ring/Ring.scala:17: secondary constructor is not allowed in universal trait extending from class Any
[error] trait Ring[@mb @sp(Byte, Short, Int, Long, Float, Double) A] extends Any with Rig[A] with Rng[A] {
```

If we remove `Any` then it will compile.

VladUreche:
That's a new bug. I'll have a look and get back asap.

tixxit:
@VladUreche I forgot about that flag! I'll try to rewrite it using the hijack. Thanks!

VladUreche:
Great! :+1: On my machine, it currently compiles well, but issues a ton of warnings -- please let me know if those are spurious and I should eliminate them.

Btw, I think you'll need to issue the `update` command by hand to load the latest plugin, for some reason sbt seems to stick to the old one by default.

Also, I couldn't get the tests to run. I tried duplicating the `laws` src to `laws-minibox` and adding it to `build.sbt`, but it produces some errors:

```
[info] Compiling 11 Scala sources to /mnt/data1/Work/Workspace/dev/algebra/laws-minibox/target/scala-2.10/classes...
[error] /mnt/data1/Work/Workspace/dev/algebra/laws-minibox/src/main/scala/algebra/laws/BaseLaws.scala:11: object std is not a member of package algebra
[error] import algebra.std.int._
[error]                ^
[error] /mnt/data1/Work/Workspace/dev/algebra/laws-minibox/src/main/scala/algebra/laws/BaseLaws.scala:33: could not find implicit value for parameter ev: algebra.PartialOrder[Int]
[error]       A.signum(x).abs ?<= 1
[error]                   ^
[error] /mnt/data1/Work/Workspace/dev/algebra/laws-minibox/src/main/scala/algebra/laws/BaseLaws.scala:36: could not find implicit value for parameter ev: algebra.Eq[Int]
[error]       A.signum(x) ?== A.sign(x).toInt
[error]               ^
[error] /mnt/data1/Work/Workspace/dev/algebra/laws-minibox/src/main/scala/algebra/laws/OrderLaws.scala:9: object std is not a member of package algebra
[error] import algebra.std.boolean._
[error]                ^
[error] /mnt/data1/Work/Workspace/dev/algebra/laws-minibox/src/main/scala/algebra/laws/OrderLaws.scala:36: could not find implicit value for parameter ev: algebra.Eq[Boolean]
[error]       A.lteqv(x, y) ?== A.gteqv(y, x)
[error]              ^
[error] /mnt/data1/Work/Workspace/dev/algebra/laws-minibox/src/main/scala/algebra/laws/OrderLaws.scala:39: could not find implicit value for parameter ev: algebra.Eq[Boolean]
[error]       A.lt(x, y) ?== (A.lteqv(x, y) && A.neqv(x, y))
[error]           ^
```

non:
Maybe `laws-minibox` isn't correctly linking to `std-minibox`? I'm not sure, I can try to look at it more tonight.

tixxit:
@non I took a rather simplistic approach to "copying" the projects. I'll fix it up tonight, don't worry.

larsrh:
Closing for being outdated.

Issue #32, opened 2015-01-11
Get Travis CI wired up


johnynek:
It would be great to see test results when reviewing.

non:
Yes! I'll try to get to this tonight or tomorrow.

PR #33, opened 2015-01-11
Topic/travis

non:
Here are some commits that should get Travis wired up (resolving #32).

I've already enabled Travis for *non/algebra* so at this point it's just a question of whether the `.travis.yml` file is  right. I based on it Spire's working file so hopefully it should be.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,10 @@
+language: scala
+scala:
+  - 2.10.4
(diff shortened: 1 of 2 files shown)


non:
Great! Looks like Travis ran and the tests passed.

johnynek:
+1 thanks!

PR #34, opened 2015-01-13
Remove Boolean, Byte, and Short specialization from most things.

### LINKED ISSUE #24: specialization cost
I am pretty dubious the specialization on Boolean and Byte is going to help at all, but it is going to increase the method count of all the traits that specialize in that way, unless I am mistaken.

The reason I don't think it is needed is that boxing a Boolean is just an if/else without allocation, and boxing a Byte is just looking up a 256 length static array (which should be in cache if you are doing this a bunch, I would expect).

It might be nice to have some kind of linter to check that we specialize all the traits the same way (just Short, Int, Long, Double, Float?).

non:
The only places where we still specialize on these types is in
Eq/PartialOrder/Order, where we specialize on everything.
This is because these type classes have much wider usage, and
also because it's likely that users will want to define instances
for case classes and other compound types which might want to
make use of specialized instances for Char, Byte, etc.

Resolves #24.


non:
I found all files I needed to modify (and also "verified" that I hadn't missed any) using `grep`. Manual review is very welcome, though obviously somewhat painful.

johnynek:
Looked at the code here and didn't notice anything off, so LGTM.

It would be so great if miniboxing became standard everywhere and we could forget all about this stuff.

non:
I'm so with you on that @johnynek. Really hate how all these `@sp` annotations look.

PR #35, opened 2015-01-16
Remove isCommutative/hasCommutative

johnynek:
My argument here is that we have two ways of specifying commutativity: 1) in the type system, 2) with a method on an instance. There is the possibility there could be badly behaved instances that don't agree with these two. Moreover, it just seems more elegant to have one way.

This replaces the methods on the instances with methods on the objects that check isInstanceOf. All the old use cases are supported, but with a new method.

In Spire, you could implement you accessors in the same way if you are wrapping an Algebra object as a Spire object, so I don't think you'll lose anything.


tixxit:
:+1:

non:
:+1:

PR #36, opened 2015-01-16
Add tests for isCommutative/isIdempotent

johnynek:
This also fixes duplicated semilattice tests by moving band and semilattice into GroupLaws.

On a side note, I'm not sure I see why we have XProperties rather than just some kind of general Properties instance. It means you can't compose inside LatticeLaws with GroupLaws. It feels a bit ad-hoc now since it is not clear where semilattice or band should go in the current picture.

I don't think it is critical, as it is just polishing the test code, but I thought I would mention my anxiety with my supportive colleagues.

### DIFF
--- a/laws/src/main/scala/algebra/laws/GroupLaws.scala
+++ b/laws/src/main/scala/algebra/laws/GroupLaws.scala
@@ -31,16 +31,23 @@ trait GroupLaws[A] extends Laws {
     Rules.repeat2("combineN", "|+|")(A.combineN)(A.combine)
   )
 
+  def band(implicit A: Band[A]) = new GroupProperties(
+    name = "band",
+    parents = List(semigroup),
+    Rules.idempotence(A.combine),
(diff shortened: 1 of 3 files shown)


non:
So I think the use of path-dependent types in Discipline means that even if we just used `Properties` we can't share them between e.g. `GroupLaws` and `LatticeLaws`. Maybe @larsrh can speak to the design there?

To some degree I wrote the `Rules` generalization to make it easier to share this stuff between sets of `Laws`. I agree it feels a bit clunky right now.

johnynek:
Got it. Not an issue, as I mentioned.

non:
:+1:

larsrh:
@johnynek That's what bases are good for – they allow you to pull in properties from a different hierarchy. I'm preparing a PR which fixes some misuses from #14 (my fault for not reviewing in time) and will also supersede this one. Sorry for the delay!

PR #39, opened 2015-01-18
Refactor laws to use bases correctly

larsrh:
Includes and supersedes #36, which suffered from a suboptimal use of bases.


non:
:+1: thanks for your help @larsrh

johnynek:
LGTM.

PR #41, opened 2015-01-22
Fix root build.sbt set up.

### LINKED ISSUE #40: root build.sbt is not inherited by projects
When I set things up I assumed the root `build.sbt` file's settings would apply to all projects. As @mpilquist points out, this isn't the case.

We should make sure all shared settings are actually shared.

non:
This change means that subprojects will inherit the root
build.sbt settings. Since this was always the intention,
this fixes the build.

Resolves #40.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,15 +1,15 @@
 // basic project info
 
-organization := "org.spire-math"
-homepage := Some(url("http://spire-math.org"))
-licenses := Seq("MIT" -> url("http://opensource.org/licenses/MIT"))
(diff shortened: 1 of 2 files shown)


non:
Review by @mpilquist, @tixxit, et al.

mpilquist:
:+1:

johnynek:
I really don't know sbt well enough to review this. Looks fine, but I am a terrible sbt cargo culter.

non:
Fair enough. @tixxit?

stew:
:+1:

PR #42, opened 2015-01-23
Make algebra.Eq[A] compatible with scala.math.Equiv[A].

non:
These changes were ported here from the 'hash' PR (#38).
This commit adds an implicit from Eq[A] => Equiv[A].
It also adds missing @sp annotations.


johnynek:
:+1:

johnynek:
yo, @tixxit @avibryant @larsrh this one is a no-brainer. Can we merge?

PR #44, opened 2015-02-07
Topic/clean up std instances

non:
Here are a bunch of new instances, some clean ups, etc.

I think all these should be good to go. However:

1. There are still some common types missing (e.g. `Vector`)
2. We need additional additive/multiplicative variants (e.g. for `Option`)
3. Maybe we should make addition a default semigroup for number types?
4. We provide `Semigroup[A] => Semigroup[Option[A]]` but `Option[_]` could use `orElse` in other cases. Worth supporting?

I don't think we have to answer all those questions immediately but we'll eventually want to.


std/src/main/scala/algebra/std/Util.scala
```diff
@@ -0,0 +1,163 @@
...
+      x.foldLeft(y) { case (z, (k, v1)) =>
+        val v3 = y.get(k) match {
+          case Some(v2) => f(v1, v2)
+          case None => v1
+        }
+        z.updated(k, v3)
+      }
+    }
+
+  def combineMap[K, V](x: Map[K, V], y: Map[K, V])(f: (V, V) => V)(g: V => V): Map[K, V] =
```
johnynek: 1) you can use the same approach is above, just reverse the order you put in the arguments to f.

2) Can you document why g is needed here but not above? Not entirely clear to me.

std/src/main/scala/algebra/std/Util.scala
```diff
@@ -0,0 +1,163 @@
...
+    y.foldLeft(x) { case (z, (k, v2)) =>
+      val v3 = x.get(k) match {
+        case Some(v1) => f(v1, v2)
+        case None => g(v2)
+      }
+      z.updated(k, v3)
+    }
+
+  def zipMap[K, V](x: Map[K, V], y: Map[K, V])(f: (V, V) => V): Map[K, V] =
```
johnynek: not clear why this is not used above. Couldn't combineMapCommutative use this?

std/src/main/scala/algebra/std/Util.scala
```diff
@@ -0,0 +1,163 @@
...
+
+    if (exponent >= 0L) loop(1L, base, exponent) else {
+      if(base == 0L) throw new ArithmeticException("zero can't be raised to negative power")
+      else if (base == 1L) 1L
+      else if (base == -1L) if ((exponent & 1L) == 0L) -1L else 1L
+      else 0L
+    }
+  }
+
+  def combineMapCommutative[K, V](x: Map[K, V], y: Map[K, V])(f: (V, V) => V): Map[K, V] =
```
johnynek: it feels a bit weird that we are not using CommutativeSemigroup here. We are giving up type safety and falling back to method name suggesting correct usage. I would prefer such a function to be private if we are doing it for efficiency reasons, and otherwise take a CommutativeSemigroup.

std/src/main/scala/algebra/std/Util.scala
```diff
@@ -0,0 +1,163 @@
...
+    if (y.size < x.size) combineMapCommutative(y, x)(f) else {
+      x.foldLeft(y) { case (z, (k, v1)) =>
```
johnynek: we found significant performance wins in creating a mutable map internally and then wrapping returning an immutable wrapper view of it (where we copy to an immutable map only on +/-). Since mostly in scalding we are aggregating these things, we just call + repeatedly.

Also not, sumOption was basically born out of this case because it is much better to allocate one mutable map and go through mutating it, then return an immutable wrapper, than to keep using the immutable.HashMap.

std/src/main/scala/algebra/std/boolean.scala
```diff
@@ -2,20 +2,40 @@ package algebra
...
 package object boolean extends BooleanInstances
 
 trait BooleanInstances {
-  implicit val booleanAlgebra = new BooleanAlgebra
+  implicit val booleanAlgebra: BooleanAlgebra =
+    new BooleanAlgebra
 }
 
-class BooleanAlgebra extends Bool[Boolean] with Order[Boolean] {
+class BooleanAlgebra extends Bool[Boolean] with Order[Boolean] with CommutativeRig[Boolean] with Serializable {
```
johnynek: maybe a comment that this is different than the CommutativeRig that is obtained from the field GF(2) (which has plus == xor).

This is a case where the most constrained structure (Field being more constrained than Rig) may not be what you are looking for.

std/src/main/scala/algebra/std/Util.scala
non: The method was written to support subtraction; this is why `g` exists, and also why the order can't easily be reversed.

Given how important these instances are (as demonstrated by your later comments) I should probably just be a bit more direct here.

std/src/main/scala/algebra/std/Util.scala
non: That's a great point -- I'll change the PR to use this strategy.

std/src/main/scala/algebra/std/byte.scala
```diff
@@ -0,0 +1,58 @@
...
+import algebra.number._
+import algebra.ring._
+import algebra.std.util.StaticMethods
+
+package object byte extends ByteInstances
+
+trait ByteInstances {
+  implicit val byteAlgebra = new ByteAlgebra
+
+  val ByteMinMaxLattice = new Lattice[Byte] {
```
johnynek: what's the thought on anonymous instances vs classes? I prefer anonymous trait style personally, but there may be scala-jvm implementation details that makes the class approach better. I'd like a rule that we know when to choose each.

johnynek: isn't this really just the `OrderLattice[Byte]` or something? Why define this for each type with on Order?

std/src/main/scala/algebra/std/char.scala
```diff
@@ -0,0 +1,19 @@
+package algebra
+package std
+
+package object char extends ByteInstances
+
+trait CharInstances {
+  implicit val charAlgebra = new CharAlgebra
```
johnynek: should we add to this or call it CharOrder?

non:
I think based on @johnynek's comments I need to go back and improve the `Map` instances a bunch.

std/src/main/scala/algebra/std/char.scala
```diff
@@ -0,0 +1,19 @@
+package algebra
+package std
+
+package object char extends ByteInstances
```
johnynek: why extend ByteInstances here?

std/src/main/scala/algebra/std/double.scala
```diff
@@ -0,0 +1,76 @@
...
+import java.lang.Double.{ longBitsToDouble, doubleToLongBits }
+import java.lang.Long.{ numberOfTrailingZeros, numberOfLeadingZeros }
+import java.lang.Math
+
+import scala.annotation.tailrec
+
+trait DoubleInstances {
+  implicit val doubleAlgebra = new DoubleAlgebra
+
+  val DoubleMinMaxLattice = new Lattice[Double] {
```
johnynek: again, don't we just want to call a function and get an `OrderLattice[Double]`?

std/src/main/scala/algebra/std/double.scala
```diff
@@ -0,0 +1,76 @@
...
+    def join(x: Double, y: Double): Double = if (x > y) x else y
+    def meet(x: Double, y: Double): Double = if (x < y) x else y
+  }
+}
+
+class DoubleAlgebra extends Field[Double] with NRoot[Double] with Order[Double] with Signed[Double] with IsReal[Double] with Serializable {
```
johnynek: should we comment here that this breaks the laws due to IEEE-754, but that it is approximately a Field?

std/src/main/scala/algebra/std/float.scala
```diff
@@ -0,0 +1,74 @@
...
+import algebra.number.{IsReal, NRoot, Signed}
+import algebra.ring.Field
+import algebra.std.util.StaticMethods
+import java.lang.Math
+import scala.annotation.tailrec
+
+trait FloatInstances {
+  implicit val floatAlgebra = new FloatAlgebra
+
+  val FloatMinMaxLattice = new Lattice[Float] {
```
johnynek: `OrderLattice[Float]`

std/src/main/scala/algebra/std/float.scala
```diff
@@ -0,0 +1,74 @@
...
+    def join(x: Float, y: Float): Float = if (x > y) x else y
+    def meet(x: Float, y: Float): Float = if (x < y) x else y
+  }
+}
+
+class FloatAlgebra extends Field[Float] with NRoot[Float] with Order[Float] with Signed[Float] with IsReal[Float] with Serializable {
```
johnynek: same comment needed to explain that this is only approximately a Field.

std/src/main/scala/algebra/std/byte.scala
tixxit: I usually prefer anonymous trait style too, but some reasons for using classes could be to allow people to extend the type class implementation or allow class annotations (eg @SerialVersionUID).

std/src/main/scala/algebra/std/list.scala
```diff
@@ -11,15 +12,21 @@ trait ListInstances {
...
-      case (Nil, _) => -1
-      case (_, Nil) => 1
-      case (x :: xs, y :: ys) =>
-        val n = ev.compare(x, y)
-        if (n != 0) n else compare(xs, ys)
-    }
+  def compare(xs: List[A], ys: List[A]): Int = {
+    @tailrec def loop(xs: List[A], ys: List[A]): Int =
+      xs match {
+        case Nil =>
```
johnynek: curios why you didn't do the double match? Does that generate worse code. I'm so sad when we have to think about this. That is what compilers should be doing.

std/src/main/scala/algebra/std/list.scala
```diff
@@ -28,7 +35,7 @@ class ListMonoid[A] extends Monoid[List[A]] {
 
   override def combineN(x: List[A], n: Int): List[A] = {
     val buf = mutable.ListBuffer.empty[A]
-      (0 until n).foreach(_ => x.foreach(buf += _))
+    (0 until n).foreach(_ => x.foreach(buf += _))
```
johnynek: if we are geeking out with perf, isn't a while loop more explicitly better? I mean, then we don't have the closure?

Again. The sadness...

std/src/main/scala/algebra/std/long.scala
```diff
@@ -0,0 +1,57 @@
...
+import algebra.lattice._
+import algebra.number._
+import algebra.ring._
+
+package object long extends LongInstances
+
+trait LongInstances {
+  implicit val longAlgebra = new LongAlgebra
+
+  val LongMinMaxLattice = new Lattice[Long] {
```
johnynek: `OrderLattice[Long]`

std/src/main/scala/algebra/std/map.scala
```diff
@@ -26,33 +50,80 @@ class MapEq[K, V](implicit ev: Eq[V], a: AdditiveMonoid[V]) extends Eq[Map[K, V]
...
+class MapMonoid[K, V](implicit sg: Semigroup[V]) extends Monoid[Map[K, V]]  {
+  def empty: Map[K, V] = Map.empty
+
+  def combine(x: Map[K, V], y: Map[K, V]): Map[K, V] =
+    StaticMethods.combineMap(x, y)(sg.combine)(v => v)
+}
+
+class MapGroup[K, V](implicit g: Group[V]) extends MapMonoid[K, V] with Group[Map[K, V]] {
+  def inverse(x: Map[K, V]): Map[K, V] =
+    x.map { case (k, v) => (k, g.inverse(v)) }
```
johnynek: might be a case for the laziness of mapValues assuming we access each item on average <= 1 times.

std/src/main/scala/algebra/std/map.scala
```diff
@@ -26,33 +50,80 @@ class MapEq[K, V](implicit ev: Eq[V], a: AdditiveMonoid[V]) extends Eq[Map[K, V]
     check(x, y) && check(y, x)
```
johnynek: why not check if the sizes are equal. If not, not equal. If they are equal we only have to `check(x, y)`

std/src/main/scala/algebra/std/list.scala
tixxit: FWIW, I benchmarked similar code a couple years ago at Precog and generally found that building a tuple and immediately matching on it was generally as fast as doing 2 matches. The code generated by Scala, at the time, was what you'd expect (allocate a Tuple2, then destructure it). Not sure if the JVM was inlining it or if stack allocation optimization kicked in. But, without benchmarks, I wouldn't want to commit to anything else here :\

std/src/main/scala/algebra/std/list.scala
tixxit: You could also allocate the closure outside of the first foreach. But, yeah, sucks that these are things we need to talk about.

tixxit: Actually, we could just use `buf ++= x` here.

std/src/main/scala/algebra/std/char.scala
tixxit: I'm guessing this was a copy/paste error. But that said, the law tests went through OK, which I think is due to us just import `algebra.std.all._` instead of individual imports. Not saying one way is better than the other, but frustrating that there isn't an easy way to verify both have all instances (without duplicating tests).

std/src/main/scala/algebra/std/map.scala
non: Yes of course. :blush:

I think maybe I was implementing an ordering and then decided to go with eq but forgot to do this.

std/src/main/scala/algebra/std/map.scala
non: Right. I'm always wary of `mapValues` due to it being a gotcha, but in this case it may make sense.

std/src/main/scala/algebra/std/list.scala
tixxit: That said, I think I'll write a quick benchmark for this...

tixxit:
Another comment I had, but left in the chat room instead, was that we need some GCD tests. Spire has a `GCDTest` that does sanity checking, but recently was forced to require Spire's `NumberTag` because the `Float`/`Double` implementations could underflow (which caused `NaN`s in LCM). I'd like to try to port the GCD tests over if possible. At least, I'd like to get some tests for the floating point (`Float` and `Double` for now) implementations, since they're a bit fiddly. I'll give it a try later today.

std/src/main/scala/algebra/std/map.scala
johnynek: scratch that. Fracking 0 == None.

std/src/main/scala/algebra/std/short.scala
```diff
@@ -0,0 +1,58 @@
...
+import algebra.number._
+import algebra.ring._
+import algebra.std.util.StaticMethods
+
+package object short extends ShortInstances
+
+trait ShortInstances {
+  implicit val shortAlgebra = new ShortAlgebra
+
+  val ShortMinMaxLattice = new Lattice[Short] {
```
johnynek: `OrderLattice[Short]`

std/src/main/scala/algebra/std/map.scala
non: So... @avibryant and @tixxit both suggested that we default to "strict" equality, meaning that `Map()` and `Map("foo" -> 0)` would not be equal.

What do you think?

johnynek: We have be burned REPEATEDLY by this question. Writing tests on Map algebras is a massive pain if you don't require normalization in the operations. But since inputs are not necessarily normalized, you have pains (so you can normalize them, but then in use who knows if they are normalized, etc...)

I think it would be good to have two variants: mapVectorEq and and mapEq. For mapVectorEq we subtract left from right and then the result has all values zero. This is the approach we finally had in tests (but Algebird has no Eq typeclass).

tixxit: :+1: We took that approach in Spire. The default was the strict version, but you could `import spire.optional.vectorEq` (or somethign similar) that would do "vector" eq/order (which was required for the law tests to pass). Of course, an alternative could be to wrap the maps in a value class or something for the vector-style operations, but that comes with another host of usability issues.

johnynek: If scala really had haskell's zero cast newtype (AnyVal feels like 1/3 the job), the wrapper approach + no orphan instances would be my preferred philosophy. Even with the weaknesses of wrappers, we use them in a few places in algebird to control typeclass selection. I don't see us doing it anywhere in algebra so far, perhaps due to the less frequent use of default typeclass instances.

std/src/main/scala/algebra/std/double.scala
non: Yes that's a good idea. I know some folks (i.e. Cats) will seat the lack of real associativity.

std/src/main/scala/algebra/std/byte.scala
non: @johnynek I chose to create non-implicit lattice instances, under the assumption that there were many possible lattices available (bitwise and/or, min/max, maybe more?).

Maybe this is the wrong strategy?

std/src/main/scala/algebra/std/map.scala
tixxit: Since all the other group operations here would be strict, I think it'd be least surprising if `inverse` was strict too.

std/src/main/scala/algebra/std/list.scala
tixxit: I wrote a benchmark for Lists only here: https://gist.github.com/tixxit/e80586e62c50bffbdfdb

Here are the results:

```
Benchmark                                        Mode  Samples          Score          Error  Units
a.b.TupleMatchBenchmark.emptyListLeft2Match     thrpt       10  293814120.445 ±  8545233.314  ops/s
a.b.TupleMatchBenchmark.emptyListLeftTuple      thrpt       10  300206447.284 ±  6360647.543  ops/s
a.b.TupleMatchBenchmark.emptyListRight2Match    thrpt       10  216331355.650 ±  5895449.200  ops/s
a.b.TupleMatchBenchmark.emptyListRightTuple     thrpt       10  301973062.006 ±  8877747.407  ops/s
a.b.TupleMatchBenchmark.emptyLists2Match        thrpt       10  376145189.382 ± 29439211.344  ops/s
a.b.TupleMatchBenchmark.emptyListsTuple         thrpt       10  135045001.983 ± 10181310.944  ops/s
a.b.TupleMatchBenchmark.nonEmptyLists2Match     thrpt       10  170827405.981 ±  5368140.256  ops/s
a.b.TupleMatchBenchmark.nonEmptyListsTuple      thrpt       10  207933353.408 ±  5254308.264  ops/s
```

The end result here is that matching on a 2-tuple lists is faster everywhere except for the `(Nil, Nil)` case. Worth looking at further. Also, at least in this case with lists, it seems that using a 2-tuple probably isn't a bad strategy.

tixxit: Just realised that @non optimized the emptyListLeft case a bit - so updating benchmark to add this implementation.

tixxit: Alright - results! I added another "Faster" case that captures the best of all worlds (ie. only destructuring if both sides are non-empty):

```
Benchmark                                              Mode  Samples          Score          Error  Units
a.b.TupleMatchBenchmark.emptyListLeft2Match           thrpt       10  288748291.733 ± 17924410.135  ops/s
a.b.TupleMatchBenchmark.emptyListLeft2MatchFast       thrpt       10  296541912.711 ± 24067230.778  ops/s
a.b.TupleMatchBenchmark.emptyListLeft2MatchFaster     thrpt       10  285169229.660 ± 39657197.426  ops/s
a.b.TupleMatchBenchmark.emptyListLeftTuple            thrpt       10  289354315.848 ± 19768765.323  ops/s
a.b.TupleMatchBenchmark.emptyListRight2Match          thrpt       10  207099008.084 ± 12826978.962  ops/s
a.b.TupleMatchBenchmark.emptyListRight2MatchFast      thrpt       10  208560165.672 ±  8428833.060  ops/s
a.b.TupleMatchBenchmark.emptyListRight2MatchFaster    thrpt       10  307506512.715 ± 22808402.454  ops/s
a.b.TupleMatchBenchmark.emptyListRightTuple           thrpt       10  301994972.708 ±  7037205.320  ops/s
a.b.TupleMatchBenchmark.emptyLists2Match              thrpt       10  401533656.442 ±  6343634.820  ops/s
a.b.TupleMatchBenchmark.emptyLists2MatchFast          thrpt       10  392072876.831 ± 40274503.858  ops/s
a.b.TupleMatchBenchmark.emptyLists2MatchFaster        thrpt       10  403947568.500 ±  4337703.883  ops/s
a.b.TupleMatchBenchmark.emptyListsTuple               thrpt       10  402480381.315 ±  6077971.326  ops/s
a.b.TupleMatchBenchmark.nonEmptyLists2Match           thrpt       10  213825113.072 ±  4839351.734  ops/s
a.b.TupleMatchBenchmark.nonEmptyLists2MatchFast       thrpt       10  215357079.103 ±  2290166.270  ops/s
a.b.TupleMatchBenchmark.nonEmptyLists2MatchFaster     thrpt       10  222770062.760 ±  7090811.748  ops/s
a.b.TupleMatchBenchmark.nonEmptyListsTuple            thrpt       10  225791825.823 ±  3292351.042  ops/s
```

tixxit: Ack - I think I've been sniped. Last benchmark, I swear! Added another benchmark which ensures the tuple is allocated on the heap (`*TupleHeap`). I've also included the bytecode scalac generates for the `checkTuple` method, just to be sure scalac wasn't doing anything fancy (it isn't). So, it seems like the JVM is taking care of the tuple strategy handily, which is nice. Results:

```
Benchmark                                              Mode  Samples          Score          Error  Units
a.b.TupleMatchBenchmark.emptyListLeft2Match           thrpt       10  253467840.436 ± 12223892.822  ops/s
a.b.TupleMatchBenchmark.emptyListLeft2MatchFast       thrpt       10  323188324.807 ± 13942236.297  ops/s
a.b.TupleMatchBenchmark.emptyListLeft2MatchFaster     thrpt       10  308222316.257 ± 17156861.142  ops/s
a.b.TupleMatchBenchmark.emptyListLeftTuple            thrpt       10  290671450.799 ± 15305844.711  ops/s
a.b.TupleMatchBenchmark.emptyListLeftTupleHeap        thrpt       10  162238249.825 ±  5178063.946  ops/s
a.b.TupleMatchBenchmark.emptyListRight2Match          thrpt       10  220221734.727 ±  3555704.452  ops/s
a.b.TupleMatchBenchmark.emptyListRight2MatchFast      thrpt       10  236875354.510 ±  4604759.625  ops/s
a.b.TupleMatchBenchmark.emptyListRight2MatchFaster    thrpt       10  332236731.746 ±  4843149.704  ops/s
a.b.TupleMatchBenchmark.emptyListRightTuple           thrpt       10  305556965.402 ±  3343750.517  ops/s
a.b.TupleMatchBenchmark.emptyListRightTupleHeap       thrpt       10  162082837.550 ±  7015805.533  ops/s
a.b.TupleMatchBenchmark.emptyLists2Match              thrpt       10  403974677.391 ±  3741407.088  ops/s
a.b.TupleMatchBenchmark.emptyLists2MatchFast          thrpt       10  374838359.717 ± 20196373.923  ops/s
a.b.TupleMatchBenchmark.emptyLists2MatchFaster        thrpt       10  382551994.617 ± 19779196.751  ops/s
a.b.TupleMatchBenchmark.emptyListsTuple               thrpt       10  383050848.374 ± 24689381.633  ops/s
a.b.TupleMatchBenchmark.emptyListsTupleHeap           thrpt       10  156116575.641 ± 10240649.596  ops/s
a.b.TupleMatchBenchmark.nonEmptyLists2Match           thrpt       10  204851392.131 ± 13503321.337  ops/s
a.b.TupleMatchBenchmark.nonEmptyLists2MatchFast       thrpt       10  199059823.003 ± 15640454.931  ops/s
a.b.TupleMatchBenchmark.nonEmptyLists2MatchFaster     thrpt       10  209463692.158 ± 12135568.592  ops/s
a.b.TupleMatchBenchmark.nonEmptyListsTuple            thrpt       10  203637089.638 ±  5339662.574  ops/s
a.b.TupleMatchBenchmark.nonEmptyListsTupleHeap        thrpt       10  126132252.642 ±  4469449.502  ops/s
```

For refrence, here is the bytecode generated by Scalac:

```
private int checkTuple(scala.collection.immutable.List<java.lang.Object>, scala.collection.immutable.List<java.lang.Object>);
  Code:
     0: new           #46                 // class scala/Tuple2
     3: dup
     4: aload_1
     5: aload_2
     6: invokespecial #50                 // Method scala/Tuple2."<init>":(Ljava/lang/Object;Ljava/lang/Object;)V
     9: astore_3
    10: aload_3
    11: ifnull        60
    14: aload_3
    15: invokevirtual #58                 // Method scala/Tuple2._1:()Ljava/lang/Object;
    18: checkcast     #60                 // class scala/collection/immutable/List
    21: astore        4
    23: aload_3
    24: invokevirtual #63                 // Method scala/Tuple2._2:()Ljava/lang/Object;
    27: checkcast     #60                 // class scala/collection/immutable/List
    30: astore        5
    32: getstatic     #69                 // Field scala/collection/immutable/Nil$.MODULE$:Lscala/collection/immutable/Nil$;
    35: aload         4
    37: invokevirtual #73                 // Method java/lang/Object.equals:(Ljava/lang/Object;)Z
    40: ifeq          60
    43: getstatic     #69                 // Field scala/collection/immutable/Nil$.MODULE$:Lscala/collection/immutable/Nil$;
    46: aload         5
    48: invokevirtual #73                 // Method java/lang/Object.equals:(Ljava/lang/Object;)Z
    51: ifeq          60
    54: iconst_0
    55: istore        6
    57: goto          199
    60: aload_3
    61: ifnull        90
    64: aload_3
    65: invokevirtual #58                 // Method scala/Tuple2._1:()Ljava/lang/Object;
    68: checkcast     #60                 // class scala/collection/immutable/List
    71: astore        7
    73: getstatic     #69                 // Field scala/collection/immutable/Nil$.MODULE$:Lscala/collection/immutable/Nil$;
    76: aload         7
    78: invokevirtual #73                 // Method java/lang/Object.equals:(Ljava/lang/Object;)Z
    81: ifeq          90
    84: iconst_m1
    85: istore        6
    87: goto          199
    90: aload_3
    91: ifnull        120
    94: aload_3
    95: invokevirtual #63                 // Method scala/Tuple2._2:()Ljava/lang/Object;
    98: checkcast     #60                 // class scala/collection/immutable/List
   101: astore        8
   103: getstatic     #69                 // Field scala/collection/immutable/Nil$.MODULE$:Lscala/collection/immutable/Nil$;
   106: aload         8
   108: invokevirtual #73                 // Method java/lang/Object.equals:(Ljava/lang/Object;)Z
   111: ifeq          120
   114: iconst_1
   115: istore        6
   117: goto          199
   120: aload_3
   121: ifnull        202
   124: aload_3
   125: invokevirtual #58                 // Method scala/Tuple2._1:()Ljava/lang/Object;
   128: checkcast     #60                 // class scala/collection/immutable/List
   131: astore        9
   133: aload_3
   134: invokevirtual #63                 // Method scala/Tuple2._2:()Ljava/lang/Object;
   137: checkcast     #60                 // class scala/collection/immutable/List
   140: astore        10
   142: aload         9
   144: instanceof    #75                 // class scala/collection/immutable/$colon$colon
   147: ifeq          202
   150: aload         9
   152: checkcast     #75                 // class scala/collection/immutable/$colon$colon
   155: astore        11
   157: aload         11
   159: invokevirtual #78                 // Method scala/collection/immutable/$colon$colon.head:()Ljava/lang/Object;
   162: invokestatic  #84                 // Method scala/runtime/BoxesRunTime.unboxToInt:(Ljava/lang/Object;)I
   165: istore        12
   167: aload         10
   169: instanceof    #75                 // class scala/collection/immutable/$colon$colon
   172: ifeq          202
   175: aload         10
   177: checkcast     #75                 // class scala/collection/immutable/$colon$colon
   180: astore        13
   182: aload         13
   184: invokevirtual #78                 // Method scala/collection/immutable/$colon$colon.head:()Ljava/lang/Object;
   187: invokestatic  #84                 // Method scala/runtime/BoxesRunTime.unboxToInt:(Ljava/lang/Object;)I
   190: istore        14
   192: iload         12
   194: iload         14
   196: isub
   197: istore        6
   199: iload         6
   201: ireturn
   202: new           #86                 // class scala/MatchError
   205: dup
   206: aload_3
   207: invokespecial #89                 // Method scala/MatchError."<init>":(Ljava/lang/Object;)V
   210: athrow
```

std/src/main/scala/algebra/std/byte.scala
johnynek: My only point is this, we seem to have a lot of the same code that is really doing this:

```scala
class OrderLattice[T](implicit ord: Order[T]) extends Lattice[T] {
  def join(x: T, y: T): T = ord.max(x, y)
  def meet(x: T, y: T): T = ord.min(x, y)
}
```
Why not write that once rather than write it for each primitive?

non:
How do people feel about this PR?

I think it would be a big improvement if we merged it. However, it was maybe too much to bite off in one PR. If folks don't feel comfortable signing-off on this, I'll close it and try to submit individual instances in a more piecemeal fashion.

I'm sorry I let this sit for so long.

/cc @tixxit @avibryant @larsrh @johnynek @ceedubs

non:
(I'm going to go through and try to address as many of @johnynek's comments as I can right now, especially when there is an obvious way to move forward.)

johnynek:
Sorry, I missed this before. Let me look tomorrow and try to merge it.

laws/src/test/scala/algebra/laws/LawTests.scala
```diff
@@ -23,6 +26,15 @@ class LawTests extends FunSuite with Discipline {
   checkAll("String", OrderLaws[String].order)
   checkAll("String", GroupLaws[String].monoid)
 
+  {
+    // TODO: test a type that has Eq but not Order
+    implicit val g: Group[Int] = Group.additive[Int]
+    checkAll("Option[Int]", OrderLaws[Option[Int]].order)
+    checkAll("Option[Int]", GroupLaws[Option[Int]].monoid)
+    checkAll("Option[String]", OrderLaws[Option[Int]].order)
```
johnynek: typo: left is not the same as right.

PS: What about a macro for some case like this so we can write the type once, and then also get a string for the type to print out.

laws/src/test/scala/algebra/laws/LawTests.scala
```diff
@@ -23,6 +26,15 @@ class LawTests extends FunSuite with Discipline {
...
+    checkAll("Option[String]", GroupLaws[Option[Int]].monoid)
```
johnynek: typo

laws/src/test/scala/algebra/laws/LawTests.scala
```diff
@@ -23,6 +26,15 @@ class LawTests extends FunSuite with Discipline {
...
+  }
+
   checkAll("List[Int]", OrderLaws[List[Int]].order)
   checkAll("List[Int]", GroupLaws[List[Int]].monoid)
   checkAll("List[String]", OrderLaws[List[Int]].order)
```
johnynek: did the other PR catch this one?

laws/src/test/scala/algebra/laws/LawTests.scala
```diff
@@ -31,13 +43,29 @@ class LawTests extends FunSuite with Discipline {
   checkAll("Set[Int]", RingLaws[Set[Int]].semiring)
   checkAll("Set[String]", RingLaws[Set[Int]].semiring)
```
johnynek: :)

laws/src/test/scala/algebra/laws/LawTests.scala
non: That's a good idea -- I'll fix the typo for now and open an issue about looking into a macro.

std/src/main/scala/algebra/std/boolean.scala
```diff
@@ -2,20 +2,45 @@ package algebra
...
 
-class BooleanAlgebra extends Bool[Boolean] with Order[Boolean] {
+/**
+ * This commutative rig is different than the one obtained from GF(2).
+ *
+ * It uses || for plus, and && for times.
+ */
+class BooleanAlgebra extends Bool[Boolean] with Order[Boolean] with CommutativeRig[Boolean] {
   def compare(x: Boolean, y: Boolean): Int =
     if (x) { if (y) 0 else 1 } else { if (y) -1 else 0 }
```
johnynek:
```scala
if (x == y) 0 else if (x) 1 else -1
```
might be my preferred way to go: for random data half the time you do only one comparison, worst case 2. Yours always does 2 (but with branch prediction who knows which would really be better).

non: Yours is terser, and the argument seems plausible. I'll change it.

std/src/main/scala/algebra/std/option.scala
```diff
@@ -0,0 +1,55 @@
+package algebra
+package std
+
+import algebra.ring._
+import scala.annotation.tailrec
+import scala.collection.mutable
+
+package object option extends OptionInstances
```
johnynek: did you want to include optionEq in here (via OptionInstances0)?

std/src/main/scala/algebra/std/map.scala
```diff
@@ -1,58 +1,132 @@
...
+
+trait MapInstances1 extends MapInstances0 {
+  implicit def mapGroup[K, V: Group]: MapGroup[K, V] =
+    new MapGroup[K, V]
+  implicit def mapAdditiveGroup[K, V: AdditiveGroup]: MapAdditiveGroup[K, V] =
+    new MapAdditiveGroup[K, V]
+}
+
+trait MapInstances0 {
+  implicit def mapEq[K, V: Eq: AdditiveMonoid]: Eq[Map[K, V]] =
```
johnynek: what about cases where we have an AdditiveMonoid, e.g. Int, but want an exact Eq (given by MapEq). I guess, in that case, you can `import std.map._` but then `implicit def mapEq[K, V: Eq]: Eq[Map[K, V]] = new MapEq`, and that will work? (does the local implicit in this case beat the imported implicit?)

johnynek:
Just the last two questions, and also not you need to merge in develop.

Thanks for doing all this work. It looks really nice to me.

std/src/main/scala/algebra/std/option.scala
non: Yes, I'll push a fix now.

std/src/main/scala/algebra/std/map.scala
non: So -- this is a specific instance of a more general problem. There are two competing interpretations of maps:

1. sparse vector (missing keys are zeros)
2. dictionary (missing keys do not exist)

I think we had agreed that we were going to standardize on (1) and force users to make an explicit import/construction to get (2). Currently you'd have to build your own local instance of `MapEq` and/or hide the "default" one.

We could also provide neither by default. This is more annoying, and affects all users, but is a lot easier to handle (bring an implicit into scope, don't worry about ambiguity).

johnynek:
LGTM.

Can we get a shipit from @tixxit who also looked at this one?

tixxit:
:+1:ship it

PR #45, opened 2015-02-09
Migrate to build.sbt and add API docs site and publishing config

travisbrown:
This is intended as a discussion PR, although I don't think there's anything terribly controversial here.

My main goal is to get the algebra API docs published in such a way that the cats API docs can automatically link to them. This is enabled on the cats side by [#158](https://github.com/non/cats/pull/158). 

Instead of just adding the sbt-unidoc and sbt-site configuration in an ad-hoc way here, I've moved the entire SBT build to follow cats in using a single `build.sbt` file instead of splitting the build config somewhat arbitrarily across multiple `build.sbt` files and `project/build.scala`.

In a couple of places I've decided to follow cats's placement of specific settings (e.g. `homepage` is set in `publishSettings` instead of across the entire build), but in general the only substantive changes have to do with the new plugins (sbt-unidoc, sbt-site, and sbt-ghpages).

There's one drive-by typo fix, and I've updated the Scala 2.11 version. These changes are unrelated to the rest of the PR. I'd be happy to decompose all of this into separate PRs if people would prefer that.


non:
:+1:

I think this is a great idea, and the unified `build.sbt` file is definitely a lot more manageable.

tixxit:
yeah - big :+1: from me - I did something similar in my miniboxing PR (clean up build, move to build.sbt), but it's a bit tied up at the moment :/

non:
Great! Merging now...

travisbrown:
Thanks much @non and @tixxit—just note that a project owner will need to [set up the `gh-pages` branch](https://github.com/sbt/sbt-ghpages) before the docs can be published for the first time.

PR #46, opened 2015-02-10
Publish API mappings metadata

travisbrown:
This is a small correction to my #45 yesterday. I'd set up automatic API mapping to work for dependencies, but not to publish the metadata for algebra itself. I've confirmed that this fix gets the right metadata into the right POMs.


ceedubs:
:+1:

non:
Thanks @ceedubs :)

non:
:+1:

PR #48, opened 2015-03-06
Eq instance creation helpers

ceedubs:
I thought these helpers might be handy. I wasn't really sure where/when specialization should be used, so I didn't use it, but I'm happy to add it in if it's desired.


non:
:+1: 

(Sorry for the long hiatus.)

ceedubs:
I'm not particularly attached to `fromUniversalEquals`. If anyone has a better name suggestion I'm happy to change it.

PR #50, opened 2015-05-08
Add Vector instances

ceedubs:
Note that the monoid for vector is not optimized.  If someone wants to suggest
any optimizations I'm happy to follow them, but mostly I just wanted to get
these instances out there.

### DIFF
--- a/std/src/main/scala/algebra/std/vector.scala
+++ b/std/src/main/scala/algebra/std/vector.scala
@@ -1,0 +1,41 @@
+package algebra
+package std
+
+import scala.annotation.tailrec
+
+package object vector extends VectorInstances
+
+trait VectorInstances {
+  implicit def vectorOrder[A: Order]: Order[Vector[A]] = new VectorOrder[A]
+  implicit def vectorMonoid[A]: Monoid[Vector[A]] = new VectorMonoid[A]
+}
+
+final class VectorOrder[A](implicit ev: Order[A]) extends Order[Vector[A]] {
+  def compare(x: Vector[A], y: Vector[A]): Int = {
+    val lengthX = x.length
+    val lengthY = y.length
+
+    @tailrec
+    def cmp(index: Int): Int = {
+      if (index == lengthX) {
+        lengthX - lengthY
+      } else if (index == lengthY) {
+        1
+      } else {
+        val n = ev.compare(x(index), y(index))
+        if (n == 0) cmp(index + 1) else n
+      }
+    }
+
+    cmp(0)
+  }
+}
+
+class VectorMonoid[A] extends Monoid[Vector[A]] {
+  def empty: Vector[A] = Vector.empty[A]
+
+  // This is really slow in 2.10: SI-7725.
+  // Unless we go all out with optimization, our implementation would probably
+  // be slower in 2.11. Do we want to optimize this?
+  def combine(x: Vector[A], y: Vector[A]): Vector[A] = x ++ y
+}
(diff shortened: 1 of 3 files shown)


std/src/main/scala/algebra/std/vector.scala
```diff
@@ -0,0 +1,41 @@
...
+import scala.annotation.tailrec
+
+package object vector extends VectorInstances
+
+trait VectorInstances {
+  implicit def vectorOrder[A: Order]: Order[Vector[A]] = new VectorOrder[A]
+  implicit def vectorMonoid[A]: Monoid[Vector[A]] = new VectorMonoid[A]
+}
+
+final class VectorOrder[A](implicit ev: Order[A]) extends Order[Vector[A]] {
```
johnynek: can you add a comment as to what the ordering is doing here? There are at least two obvious ones. Looks like this one is lexicographic order.

std/src/main/scala/algebra/std/vector.scala
```diff
@@ -0,0 +1,41 @@
...
+        val n = ev.compare(x(index), y(index))
+        if (n == 0) cmp(index + 1) else n
+      }
+    }
+
+    cmp(0)
+  }
+}
+
+class VectorMonoid[A] extends Monoid[Vector[A]] {
```
johnynek: can we comment here as to what the monoid is and name it something like, VectorFreeMonoid or VectorConcatMonoid?

std/src/main/scala/algebra/std/vector.scala
```diff
@@ -0,0 +1,41 @@
...
+  def empty: Vector[A] = Vector.empty[A]
+
+  // This is really slow in 2.10: SI-7725.
+  // Unless we go all out with optimization, our implementation would probably
+  // be slower in 2.11. Do we want to optimize this?
```
johnynek: by implementing combineAll and combineAllOption you can mitigate this for the case of combining many things (using a Vector.newBuilder). All free-like Monoids will have this issue of O(N^2) cost to combine N items.

std/src/main/scala/algebra/std/vector.scala
non: Agreed. Also it would be a good idea to figure out if we want to endorse this as the *the* `Monoid[Vector[A]]` instance (since it is parametric) or merely as an alternative to `Monoid[A] => Monoid[Vector[A]]` (since that is what folks treating vectors as vectors probably want).

johnynek: Yes, I overlooked that import vector._ would give you this one. In algebird we made `Monoid[List[T]] ` the free monoid here, but we used `Monoid[T] => Monoid[Vector[T]]` to do pointwise addition (with padding of zeros to the longer vector, but what we really want is the length as a type parameter and only add vectors of the same length, but we didn't geek out with type-level Nats).

std/src/main/scala/algebra/std/vector.scala
ceedubs: Sure. I used lexicographic order because I saw that's what the Order[List[A]] implementation was doing. Comparing vector length first would certainly be more efficient. I don't really see myself ever using either, so if others have opinions on what this should be doing, I'm happy to follow them.

This also makes me realize that if we take this approach, I should override `eqv` to first check that the lengths are equal.

ceedubs:
I've neglected this for a long time, and I think that it probably doesn't provide the most useful `Vector` instances for the sorts of things that a lot of algebra users will want to use vectors for. I'm going to go ahead and close this out so it's not hanging around the pull request list giving false impressions of being active. Anyone who wants to work on this can of course take any of the code from this that may be helpful.

PR #51, opened 2015-05-11
Fix Eq derivative method

pjan:
Believe a small typo crept in...


ceedubs:
Yikes. Good catch! :+1:

It would probably be good to have a unit test that calling `on` with the identity function results in an `Eq` instance that produces the same result for `eqv(x, y)`. That should be fairly simple to write and should have caught this bug. (On this note, the Cats contravariant functor tests probably also should have caught this - we should see if we can make that happen.)

non:
:+1:

PR #54, opened 2015-05-13
use simlacrum everywhere possible

pjan:


johnynek:
I'm a bit on the fence on this one. The win is positive, but not overwhelming (200 lines added 300 removed), I'm not sure about the future of macro-annotations (but maybe that is clear), and lastly, it adds a learning curve to contributing (for instance, I had to read the source of simlacrum to see what noop did as it is not mentioned on the front page and there is no wiki).

non:
@johnynek OK. I think the documentation will probably improve, but I am happy to close this for now and re-open the conversation once that happens.

non:
@pjan Thanks for doing this, but I think we will be deferring this issue for now. Is it OK if I close this PR?

pjan:
@johnynek it's unfortunate that simulacrum has not yet managed to create the documentation, which would lower its barrier for usage/learning curve (which is not really high). I very much appreciate the code clarity and consistent behaviour it brings (taking away the need to create typeclass boilerplate stuff), yet de gustibus... .

@non PR was only made to take some work of your shoulders. No offence taken if it is not being merged. Will likely put up a small one with some tiny code improvements I came across when making this one.

non:
Thanks! I agree that in the long run I think something like Simulacrum is the right thing. But Oscar is also right that there isn't enough documentation right now.

antonkulaga:
The main problem with simulacrum is that Intellij cannot regognize the code generated by it, I opened an issue about that ( https://youtrack.jetbrains.com/issue/SCL-8800 ) but I doubt that it will be fixed soon.

denisrosset:
What about the speed? Does the code generated by simulacrum inline the method calls, as machinist does in spire?

Second, what are the binary compatibility guarantees of simulacrum?

sritchie:
Shall we close this for now and pick it back up down the road?

larsrh:
Closing for being outdated.

PR #60, opened 2015-06-09
OrderLaws s/eq/eqv

adelbertc:


non:
This looks good. Thanks!

The background here is that it is very difficult to call a method named `eq`.

:+1:

non:
Since we discussed using N=1 as our new PR review policy, I am merging this now.

PR #62, opened 2015-06-10
PartialOrder instance for Set

adelbertc:


std/src/main/scala/algebra/std/set.scala
```diff
@@ -6,12 +6,15 @@ import algebra.ring.Semiring
...
   implicit def setSemiring[A] = new SetSemiring[A]
 }
 
-class SetEq[A] extends Eq[Set[A]] {
-  def eqv(x: Set[A], y: Set[A]): Boolean = x == y
+class SetPartialOrder[A] extends PartialOrder[Set[A]] {
+  def partialCompare(x: Set[A], y: Set[A]): Double =
+    if (x == y) 0.0
+    else if (x.subsetOf(y)) -1.0
+    else 1.0
```
johnynek: I think you want `if (y.subsetOf(x)) 1.0 else java.lang.Double.NaN` right?

johnynek:
need to add in that this follows the PartialOrder laws.

std/src/main/scala/algebra/std/set.scala
```diff
@@ -6,12 +6,15 @@ import algebra.ring.Semiring
...
 
 trait SetInstances {
-  implicit def setEq[A]: Eq[Set[A]] = new SetEq[A]
+  implicit def setPartialOrder[A]: PartialOrder[Set[A]] = new SetPartialOrder[A]
   implicit def setSemiring[A] = new SetSemiring[A]
 }
 
-class SetEq[A] extends Eq[Set[A]] {
-  def eqv(x: Set[A], y: Set[A]): Boolean = x == y
+class SetPartialOrder[A] extends PartialOrder[Set[A]] {
```
johnynek: might want to override eqv so it will not have to do two subset checks when they are not equal.

johnynek:
Also, while we're at it, I think you can make a `Lattice` instance for `Set[A]` for which meet is intersection and join is union (or the other way around), which is basically the same as the Semiring instance, but it also follows the Lattice laws.

adelbertc:
Currently there doesn't seem to be any tests in the project - I am more than happy to start things off by writing the tests for `Set` though.

Will address your comments in an incoming commit.

johnynek:
See the laws sub project. It depends on std.

adelbertc:
Doh, silly me. Will add, thanks!

std/src/main/scala/algebra/std/set.scala
adelbertc: `y.subsetOf(x)` makes sense to me, but the `else Double.NaN` would break `PartialOrder` laws, where

`x > y --> y < x`

Looking at: https://github.com/non/algebra/blob/master/core/src/main/scala/algebra/PartialOrder.scala#L111

Given sets `x = Set(0, 1), y = Set(0)`, `x > y` yields `true` but `y < x` yields `Double.NaN`.

std/src/main/scala/algebra/std/set.scala
```diff
@@ -1,17 +1,31 @@
...
 
-class SetEq[A] extends Eq[Set[A]] {
-  def eqv(x: Set[A], y: Set[A]): Boolean = x == y
+class SetLattice[A] extends Lattice[Set[A]] {
+  def join(lhs: Set[A], rhs: Set[A]): Set[A] = lhs.union(rhs)
+
+  def meet(lhs: Set[A], rhs: Set[A]): Set[A] = lhs.intersect(rhs)
+}
+
+class SetPartialOrder[A] extends PartialOrder[Set[A]] {
```
non: Would you consider something that makes fewer passes over the sets? For example:

```scala
def partialCompare(x: Set[A], y: Set[A]): Double =
  if (x.size < y.size) (if (x.forall(y)) -1.0 else Double.NaN)
  else if (y.size < x.size) -partialCompare(y, x)
  else if (x == y) 0.0
  else Double.NaN
```

It's a bit junkier, but in the worst case you are doing one iteration over the smaller of the two sets instead of potentially three.

adelbertc: Good point, I was wondering if I was making too many passes over the `Set`s - change has been made.

non:
:+1:

johnynek:
Nice work! Thanks.

PR #65, opened 2015-06-12
Stay DRY between tested types and test names.

### LINKED ISSUE #57: Use macros to remove duplicate between tests and names
Our tests have a lot of lines like these:

```scala
checkAll("List[Int]", OrderLaws[List[Int]].order)
checkAll("List[Int]", GroupLaws[List[Int]].monoid)
checkAll("List[String]", OrderLaws[List[String]].order)
checkAll("List[String]", GroupLaws[List[String]].monoid)
```

In addition to being annoyingly repetitive these calls introduce the very real possibility that the label and the actual code being run can be in disagrement.

It would be nice to use a macro (or similar) to remove the duplication.

TomasMikula:
The solution uses implicit TypeTags to obtain type names.

Resolves #57.

There is a slight change in the name of the tested property. Where the property name used to be

    Rat.field.additive:commutative group.consistent subtract

it is now

    [algebra.std.Rat].field.additive:commutative group.consistent subtract

since the type name obtained from TypeTag is a fully qualified name.

### DIFF
--- a/laws/src/test/scala/algebra/laws/LawTests.scala
+++ b/laws/src/test/scala/algebra/laws/LawTests.scala
@@ -1,6 +1,8 @@
 package algebra
 package laws
 
+import scala.reflect.runtime.universe.TypeTag
+
 import algebra.lattice._
 import algebra.number._
 import algebra.ring._
@@ -9,7 +11,9 @@ import algebra.std.Rat
 
 import CheckSupport._
 
+import org.typelevel.discipline.{Laws, Predicate}
 import org.typelevel.discipline.scalatest.Discipline
+import org.scalacheck.Arbitrary
 import org.scalatest.FunSuite
 
 class LawTests extends FunSuite with Discipline {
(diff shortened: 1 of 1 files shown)


non:
This looks really nice, thanks!

:+1: from me for sure.

johnynek:
Awesome!

larsrh:
@TomasMikula If you're inclined, feel free to file a pull request to discipline!

TomasMikula:
@larsrh I had to roll my own version of discipline for a project to overcome some limitations/complexities of discipline. Hopefully I will get around to bringing up that discussion soon.

larsrh:
Would love to hear about those.

TomasMikula:
Alright, I compelled myself to write it up and publish it:
https://github.com/TomasMikula/Principled#comparison-to-discipline
Interested in your feedback.

Also, I have no interest in fragmenting the ecosystem, so it would be cool if discipline could adopt that approach (or address the issues in any other way).

non:
@TomasMikula Thanks for doing this!

I can confirm that some of the pain points you found are things we have worked around too. Also, I think something like Cats' `IsEq[A]` type (for deferring/abstracting over equivalence/equality) would be really nice to have in Discipline.

PR #66, opened 2015-07-01
Fix typo in README

YawarRaza7349:


johnynek:
:+1:

johnynek:
We are still at N = 2, right? (at least the readme says so). @non ?

non:
Yes! I think we agreed to change N to 1 but haven't updated the README.

I'll merge this and then look at that.

Issue #69, opened 2015-07-07
Add byAll and fromAll to Eq/PartialOrder/Order


non:
Something that I've received requests for is an easier way to define these instances.

Here's an example of what I mean (assume we can't automatically derive instances for case classes/products):

```scala
case class Foo(a: Int, b: String, c: Int)

// a function is used if the previous function returned 0/EQ
val o1: Order[Foo] = Order.fromAll[Foo](
  (x, y) => x.a compare y.a,
  (x, y) => x.b compare y.b,
  (x, y) => x.c compare y.c)

// here's an even more terse version that will use Order[Int] and Order[String]
val o2: Order[Foo] = Order.byAll(_.a, _.b, _.c)
```

The logic in both cases is that the first function is used -- and if the items are equal, the next one is used, otherwise the result returns immediately. If after the last function runs they are equal, then 0/EQ is returned.

johnynek:
we've been making macros for these cases. I assume the macros offer better performance, but maybe I'm wrong. Certainly better usability.

I can see adding these "generalization over case class" instances being something we have standard macros for in many of our cases: Semigroup, Monoid, Group, Ring (but not Field), Order, Eq, PartialOrder, and probably more.

non:
I would be happy to use macros. I agree that the performance will be much better when function literals are used (they can be inlined into the code you would explicitly write, more-or-less).

johnynek:
I bet if we are careful, we can get a lot of code reuse between the different typeclasses over case-classes.

PR #70, opened 2015-07-08
Added fromAll to Eq,Order,PartialOrder (#69)

FrankRaiser:


non:
Thanks, this looks good!

@johnynek would you prefer to let this PR sit awhile and while we look into a macro solution? Or should we merge this and then replace it with a macro?

core/src/main/scala/algebra/Order.scala
```diff
@@ -143,6 +143,16 @@ object Order extends OrderFunctions {
     }
 
   /**
+   * Define an `Order[A]` using the given functions `fs` until
+   * the first one returns a non-0 result.
+   */
+  def fromAll[@sp A](fs: ((A, A) => Int)*): Order[A] =
+    new Order[A] {
+      def compare(x : A, y: A) =
+        fs.map(_(x,y)).dropWhile(_ == 0).headOption.getOrElse(0)
```
johnynek: this will evaluate all of them. What about:

```scala
fs.foldLeft(0) {
  case (0, fn) => fn(x, y)
  case (r, _) => r
}
```

core/src/main/scala/algebra/PartialOrder.scala
```diff
@@ -158,6 +158,16 @@ object PartialOrder extends PartialOrderFunctions {
     }
 
   /**
+   * Define a `PartialOrder[A]` using the given functions `fs` until
+   * the first one returns a non-0 result or NaN.
+   */
+  def fromAll[@sp A](fs: ((A, A) => Double)*): PartialOrder[A] =
+    new PartialOrder[A] {
+      def partialCompare(x : A, y: A) =
+        fs.map(_(x,y)).dropWhile(_ == 0).headOption.getOrElse(0.0)
```
johnynek: same comment about avoiding total evaluation using foldLeft

core/src/main/scala/algebra/PartialOrder.scala
```diff
@@ -158,6 +158,16 @@ object PartialOrder extends PartialOrderFunctions {
     }
 
   /**
+   * Define a `PartialOrder[A]` using the given functions `fs` until
+   * the first one returns a non-0 result or NaN.
+   */
+  def fromAll[@sp A](fs: ((A, A) => Double)*): PartialOrder[A] =
```
johnynek: maybe `fromFirst` here? Because as soon as you hit a difference you choose it, right? Not really all? Same comment above.

core/src/main/scala/algebra/Eq.scala
```diff
@@ -69,6 +69,11 @@ object Eq extends EqFunctions {
       def eqv(x: A, y: A) = f(x, y)
     }
 
+  def fromAll[A](fs : ((A,A) => Boolean)*) : Eq[A] =
```
johnynek: you are `@sp` below, but not here.

johnynek: That said, I doubt we need `@sp`. You'd almost never want it here right? This is for classes/traits I assume.

johnynek:
I'm happy to merge this after some comments. It could be useful in cases the macros could not easily cover (traits, non-case-classes).

johnynek:
style note: inconsistent spacing with `:`. Prefer `(a: A): B` style to `(a : A) : B`

core/src/main/scala/algebra/PartialOrder.scala
FrankRaiser: fromFirst was the name @non used for the ticket, but I agree. In fact, I see the same problem with Order as well. Both Order and PartialOrder really determine the result via the first function that defines any order between the elements. Then again, I can equally accept reading it as "constructing an order derived from all of these functions used in turn". I'm impartial on fromAll or fromFirst, but I think it should be the same for Order and PartialOrder then.

core/src/main/scala/algebra/Eq.scala
```diff
@@ -70,6 +70,19 @@ object Eq extends EqFunctions {
     }
 
   /**
+   * Create an `Eq` instance that defines equality via the conjunction
+   * of each of the given functions `fs`.
+   */
+  def fromAll[@sp A](fs : ((A,A) => Boolean)*) : Eq[A] =
+    new Eq[A] {
+      def eqv(x: A, y: A) =
+        fs.foldLeft(true) {
```
johnynek: is this what we intend? This is different than the logic you had before (the others below are not).

Here you are looking for the first true, before you were keeping if all were true. I think we want the all true case.

johnynek:
@non what do you think about `@sp` on these methods? I think we don't need it and it will cause us some jar bloat. These methods don't really make sense on primitives.

non:
Yes, I agree. These generic type parameters should not be specialized.

johnynek:
@non and you prefer fromAll even in the lexicographic ordering?

non:
I'm not sure. Mostly I wanted a name to relate it to `from` but to make it clear it uses several functions.

core/src/main/scala/algebra/Eq.scala
FrankRaiser: I believe this behaves as documented. If any of the method returns false, the fold stops with false as a result. The only way to get true would be for all methods to return true. I'm not sure why you think this only looks for the first true.
I wrote another test to verify this behavior, but I'm not sure if I should add that, since it is a standard exemplary test, not a law test.

johnynek:
@non I like the `from` part, it is just that the `all` part does not sound lexicographic to me. `fromLazily`?

non:
Sure, `fromLazily` seems good to me.

johnynek:
@FrankRaiser I think we got consensus on `fromLazily` for the Order case, if you think that sounds okay too.

FrankRaiser:
Sounds fine with me. I'm a bit short on time now, but should get an updated PR ready this weekend.

johnynek:
Cool. Thanks for the help!

johnynek:
Looks good to me. If we can make this merge cleanly, I'll click the shiny green button.

johnynek:
@FrankRaiser will you be able to get a clean merge here? It would be great to merge any binary incompatible change as soon as possible (this would be one).

johnynek:
@non about my previous comment: maybe these constructor functions should go into a package like std? They are not changing the shape of the traits, but I *think* (could be wrong) that they will change the binary (adding methods to traits causes issues, maybe not to objects).

laws/src/test/scala/algebra/laws/EqTest.scala
```diff
@@ -0,0 +1,29 @@
...
+import java.util
+
+import algebra.Eq
+import org.scalatest._
+import algebra.std.all._
+
+/**
+ * @author Frank Raiser <raiser.frank@gmail.com>
+ */
+class EqTest extends FlatSpec with Matchers {
```
johnynek: Isn't this a subset of the property check below?

johnynek:
we moved `Eq` `Order` etc.. to cats-kernel in typelevel/cats.

PR #72, opened 2015-07-18
Should improve the workflow when releasing to Sonatype.

non:
This commit adds sbt-sonatype. It also updates our test
library dependencies to the most recent versions and
cleans up the SBT config a bit.


non:
NOTE: I haven't actually published algebra with this setup yet, but I've used a very similar set up to publish at least 4 other projects today, so I'm fairly confident it will work.

non:
P.S. I learned the hard way that the sort of setup we already have (with sbt-release) does not necessarily work well with sbt-scalajs: https://github.com/InTheNow/sbt-scalajs/issues/19

tixxit:
:+1:

InTheNow:
Also, discipline is not released, despite https://github.com/typelevel/discipline/pull/13

non:
This PR is subsumed by #73.

PR #73, opened 2015-07-22
Add scala.js support

InTheNow:
Adds sjs support


non:
Thanks! This looks really good.

What are the `OBSOLETE` files all about? Should they be removed?

InTheNow:
Yes. I just chatted on gitter, they should go. I'll wait for any other review feedback and fix in one go

tixxit:
:clap:

InTheNow:
Also, I should add `sbt-sonatype`, effectively merging in https://github.com/non/algebra/pull/72

non:
@tixxit signed off on that one, so I'm just going to merge it to master. Then you can merge it here.

non:
So here are some review comments:

1. Rather than `RInt.apply` I'd prefer something like `Platform.rint` to make it a bit clearer what is going on. That would also mean we could put future JS/JVM split code into `Platform` too.

2. What was the rationale for changing `TypeTag` to `TypeTagM`? Are there problems with type tags in scala.js? Was there a 2.10/2.11 compatibility problem we hadn't noticed? Something else?

Other than that, this looks great. Thanks!

InTheNow:
1. Name change is fine. `rint` will be fixed in next sjs release ( probably by me) - https://github.com/scala-js/scala-js/issues/1797, but a general place holder such as Platform is  a good idea eg - [scalacheck has one too, and better still, private](https://github.com/rickynils/scalacheck/blob/master/jvm/src/main/scala/org/scalacheck/Platform.scala)

2. `TypeTag` is runtime reflection, so is not/will not be in scala.js, hence the macro implementation.

InTheNow:
Also, there are a few dangling white spaces that I'll remove, and some documentation for the macros, platform diffs and so on

InTheNow:
Done - also moved `SerializableRules` to `Platform`

non:
I'm not sure why Travis failed, but this looks good to me. :+1:

InTheNow:
Sbt has fixed their URL issue, can someone with access re-run the tests, please?

non:
Tests pass, look great. :+1:

PR #76, opened 2015-08-02
Add macros to released artifacts

InTheNow:
Macros artifact is missing from the release. Please can you also do a new release with this ASAP, please? 

The good news is that cats is almost ported - that's how I found this, and have tested locally.


non:
:+1:

PR #77, opened 2015-08-02
Update the README to 0.3.1.

non:


johnynek:
shouldn't we just do this when we publish the jar in the same step (i.e. when we set the version for the jar, etc)? I mean, there is no need to make reviewing a change like this a higher bar than the publishing itself.

non:
That's a good point. In the future I'll just go ahead and do that.

PR #80, opened 2015-08-22
Typo fixes in comments

pleira:


non:
Good catch, thanks! :+1:

non:
I'm going to wait for another +1. In Cats we have a policy that documentation-only fixes (especially ones like this that involve fixing typos, etc) can be merged immediately. How do folks here feel about that?

johnynek:
:+1: and :+1: to doc-only changes not needing a second review (honestly, I thought we moved to one +1 for this project for everything).

non:
You're right, we did, I had forgotten! :blush:

PR #87, opened 2015-10-08
Add Unit instances

ceedubs:
Mostly I wanted the `Order` instance for `Unit`. The use-case is that in
some tests we want to check if something two instances of
`Error Xor Unit` are equal, and the derived `Eq` instance for `Xor`
needs an `Eq` instance for both sides.

I have no idea whether the `CommutativeRig[Unit]` is a good idea or is
useful. It seems to be law-abiding though. Let me know if you want me to
rip it out.

### DIFF
--- a/std/shared/src/main/scala/algebra/std/unit.scala
+++ b/std/shared/src/main/scala/algebra/std/unit.scala
@@ -21,7 +21,7 @@ class UnitAlgebra extends Order[Unit] with CommutativeRing[Unit] with BoundedSem
   override def lteqv(x: Unit, y: Unit): Boolean = true
 
   override def min(x: Unit, y: Unit): Unit = ()
-  override def max(x: Unit, y: Unit): Unit = Unit
+  override def max(x: Unit, y: Unit): Unit = ()
 
   def zero: Unit = ()
   def one: Unit = ()
--- a/std/shared/src/main/scala/algebra/std/unit.scala
+++ b/std/shared/src/main/scala/algebra/std/unit.scala
@@ -1,16 +1,16 @@
 package algebra
 package std
 
-import algebra.ring.CommutativeRig
+import algebra.ring.CommutativeRing
 
 package object unit extends UnitInstances
 
 trait UnitInstances {
-  implicit val unitAlgebra: Order[Unit] with CommutativeRig[Unit] =
+  implicit val unitAlgebra: Order[Unit] with CommutativeRing[Unit] with BoundedSemilattice[Unit] =
     new UnitAlgebra
 }
 
-class UnitAlgebra extends Order[Unit] with CommutativeRig[Unit] {
+class UnitAlgebra extends Order[Unit] with CommutativeRing[Unit] with BoundedSemilattice[Unit] {
   def compare(x: Unit, y: Unit): Int = 0
 
   override def eqv(x: Unit, y: Unit): Boolean = true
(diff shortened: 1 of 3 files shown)


non:
I'm pretty sure you can upgrade it to a `CommutativeRing[Unit]` with no problems (it is a *zero ring*: https://en.wikipedia.org/wiki/Zero_ring). There's little harm in having these instances, even if they are mostly just a novelty.

johnynek:
+1 to Ring. Also I think you get Lattice here.

johnynek:
actually, `BoundedSemilattice` is what we I should have said instead of `Lattice`.

ceedubs:
@non @johnynek thanks for the fast feedback and suggestions. I've made the unit algebra extend `CommutativeRing` and `BoundedSemilattice`.

Am I doing the right law-checking?  It's straightforward for `order` and `boundedSemilattice`, but I wasn't sure what to do for `CommutativeRing`. I didn't spot a `def commutativeRing` in `RingLaws` that would have been the obvious answer.

ceedubs:
Also thanks for the link @non. That was helpful.

std/shared/src/main/scala/algebra/std/unit.scala
```diff
@@ -0,0 +1,39 @@
...
+
+  override def eqv(x: Unit, y: Unit): Boolean = true
+  override def neqv(x: Unit, y: Unit): Boolean = false
+  override def gt(x: Unit, y: Unit): Boolean = false
+  override def lt(x: Unit, y: Unit): Boolean = false
+  override def gteqv(x: Unit, y: Unit): Boolean = true
+  override def lteqv(x: Unit, y: Unit): Boolean = true
+
+  override def min(x: Unit, y: Unit): Unit = ()
+  override def max(x: Unit, y: Unit): Unit = Unit
```
non: Shouldn't this be `()` ?

ceedubs: Oops good catch. I had forgotten that this would even compile in Scala.

ceedubs: Fixed.

non:
:+1:

tixxit:
I don't think we should be able to get anything beyond `Rng` here. I think a Ring implies that 1 and 0 are different. OTOH, the laws seem to pass...

tixxit:
Ah - nevermind - this ring is the only ring where 1 != 0! Carry on :)

std/shared/src/main/scala/algebra/std/unit.scala
```diff
@@ -0,0 +1,39 @@
...
+  override def isZero(x: Unit)(implicit ev: Eq[Unit]): Boolean = true
+  override def isOne(x: Unit)(implicit ev: Eq[Unit]): Boolean = true
+
+  def plus(a: Unit, b: Unit): Unit = ()
+  def negate(x: Unit): Unit = ()
+  def times(a: Unit, b: Unit): Unit = ()
+  override def pow(a: Unit, b: Int): Unit = ()
+
+  def empty: Unit = ()
+  def combine(x: Unit, y: Unit): Unit = ()
```
johnynek: small note, `combineAllOption` could be implemented to not bother to iterate, but maybe that's getting silly since it is hard to imagine this being a performance issue.

ceedubs: Maybe it's getting silly, but I think it's a good point. I'm on a different computer at the moment. I'll try to remember to come back to this.

PR #88, opened 2015-10-08
Add Mima for core

johnynek:
This adds mima for core and makes it a build failure if compatibility is broken.

We can imagine being stricter and doing the same for std. Should we do that?

Before algebird can depend on this, I think we need a very strict policy that we only break compatibility if we find real flaws. The cost to changing binaries on a library this low in the stack is pretty heavy since it could wind up being a major diamond dependency.

PS: I wish we could somehow solve the diamond dependency problem (private dependencies, versioned imports, something...)


non:
:+1:

I think `core` is pretty stable. Before we add this to `std` I'd prefer to really get to the crux of what we expect from it (for example, I think it might be nice to remove `Rat`).

PR #91, opened 2015-10-09
Override repeatedCombineN and combineAllOption for Unit algebra

ceedubs:
The `combineAllOption` override is per a [suggestion](https://github.com/non/algebra/pull/87#discussion_r41585052) from @johnynek.


johnynek:
Thanks. We have the algebra for Unit totally dialed. I guess Boolean is close too. Next we need to start working on the type with 3 elements.

ceedubs:
:)

Issue #89, opened 2015-10-08
Heyting#asCommutativeRing broken


TomasMikula:
`Heyting#asCommutativeRing` seems to be broken. The problem is with the [`negate`](https://github.com/non/algebra/blob/0f8f8df5b62473be5bea06266747657b56ff196a/core/src/main/scala/algebra/lattice/Heyting.scala#L58) method: `complement` is not an inverse with respect to `xor`.

As an example, consider the usual interpretation of the type `Boolean = {false, true}` as a Boolean algebra.

From the additive group, we have the law

    plus(x, negate(x)) = zero

Substituting `one` for `x` we get

    plus(one, negate(one)) = zero

And now replace `plus` by `xor`, `negate` by `¬`, `zero` by `false` and `one` by `true` (as defined in `asCommutativeRing`):

    xor(true, ¬true) = false

i.e.

    xor(true, false) = false

which is a contradiction.

FWIW, I don't immediately see how this could be fixed, i.e. I don't see how every Heyting algebra gives rise to a commutative ring.

non:
Oops! You are right that this implementation is buggy.

I think that in this case, every element is its own inverse, since `xor(x, x) = false` for any `x`. In that case, `negate(x) = x`.

non:
I'll try to prepare a patch with laws, law-checking, and a correction. If I can't prove (and test) that this change is correct, we'll have to remove the method.

TomasMikula:
That seems to work.

TomasMikula:
Actually, I'm not sure. I realized that I just checked that for a Boolean algebra, not Heyting algebra.

non:
From `Heyting`:

```scala
def xor(a: A, b: A): A = or(and(a, complement(b)), and(complement(a), b))

xor(a, a)
// expand definition
or(and(a, complement(a)), and(complement(a), a))
// we know that and(a, complement(a)) is false even in just heyting algebras
or(false, false)
// we know this is false
false
```

So I'm pretty sure we know `xor(a, a) = false`.

TomasMikula:
The tricky part IMO will be to prove the ring's distributive law.

TomasMikula:
Here's why it cannot be true. Suppose the object defined by
Heyting#asCommutativeRing is a ring. Then it is a ring with a property a.a
= a. A ring with this property is called a Boolean ring and is equivalent
to a Boolean algebra (with xor defined as ring addition). Since we know
[...]

PR #94, opened 2015-10-12
Remove asCommutativeRing from Heyting

johnynek:
Addresses the issue in #89

### DIFF
--- a/core/src/main/scala/algebra/lattice/Bool.scala
+++ b/core/src/main/scala/algebra/lattice/Bool.scala
@@ -1,31 +1,45 @@
 package algebra
 package lattice
 
+import ring.CommutativeRing
 import scala.{specialized => sp}
 
 /**
  * Boolean algebras are Heyting algebras with the additional
  * constraint that the law of the excluded middle is true
  * (equivalently, double-negation is true).
- * 
+ *
  * This means that in addition to the laws Heyting algebras obey,
  * boolean algebras also obey the following:
- * 
+ *
  *  - (a ∨ ¬a) = 1
  *  - ¬¬a = a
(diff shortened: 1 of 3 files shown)


non:
:+1:

core/src/main/scala/algebra/lattice/Bool.scala
```diff
@@ -1,31 +1,45 @@
...
+   * Every Boolean algebra is a CommutativeRing, but we don't extend
+   * CommutativeRing because, e.g. we might want a Bool[Int] and CommutativeRing[Int] to
+   * refer to different structures, by default.
+   */
+  def asCommutativeRing: CommutativeRing[A] =
+    new CommutativeRing[A] {
+      def zero: A = self.zero
+      def one: A = self.one
+      def plus(x: A, y: A): A = self.xor(x, y)
+      def negate(x: A): A = self.complement(x)
```
TomasMikula: `negate` here is still broken. It should be

```scala
def negate(x: A): A = x
```

non: Aha, right. I'll fix that now. Thanks!

TomasMikula:
Also note that if we had a type `BoolRing`, it could implement `negate` (to be `negate(x) = x`) and then the instance returned from `Bool#asCommutativeRing` would not have to override `negate`.

non:
Let's open an issue about possibly adding `BoolRing`, and I will submit a PR to fix this bug. (I also want to ensure that we have a test that detects this issue.)

TomasMikula:
Sounds good, I will open the issue.

PR #96, opened 2015-10-12
Fix band specialization

### LINKED ISSUE #90: Band is specialized on too much
https://github.com/non/algebra/blob/master/core/src/main/scala/algebra/Band.scala#L10

we restricted later (to avoid code bloat) to `Int, Long, Float, Double`. Looks like we missed it on band. What should we do?

Dropping it will break binary compatibility, so we should decide soon. It would be nice to have a way to set this in one place or an easy way to grep for the specializations. There may be others we missed.

johnynek:
fixes #90 

Question: should Band and Semilattice use just Int, Long? Looks like most or all of algebra.lattice does this (even though you could make a lattice of Doubles/Floats with max/min).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -61,7 +61,7 @@ lazy val aggregate = project.in(file("."))
 lazy val core = crossProject.crossType(CrossType.Pure)
   .settings(moduleName := "algebra")
   .settings(mimaDefaultSettings: _*)
-  .settings(previousArtifact := Some("org.spire-math" %% "algebra" % "0.3.1"))
+  .settings(previousArtifact := None)
   .settings(algebraSettings: _*)
   .settings(sourceGenerators in Compile <+= (sourceManaged in Compile).map(Boilerplate.gen))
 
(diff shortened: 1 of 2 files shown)


johnynek:
in my opinion, keeping Float/Double makes as much sense as almost any algebraic structure (except maybe we could say that Ring/Field would be by far the most common, so maybe only they should be specialized on Double, Float?).

non:
I guess I am loathe to remove `Float` / `Double` specialization unless we are sure they won't (or can't) be used. Do you think it would be OK to leave these annotations on our type classes? Given that Spire currently does specialize these, I feel like continuing to do that is a conservative approach that is consistent with enforcing MiMA.

johnynek:
Yeah, I'm totally fine with leaving Float/Double, but I just noticed that our algebra.lattice packages don't have it, and I'm not sure why.

non:
That was probably just an oversight on my part. Let's add them.

johnynek:
@non can we do that in a separate fix? I like to keep them focused. If so, are there any other things we can add here?

non:
Sure thing. This looks good to me. :+1:

PR #97, opened 2015-10-14
Fix bug in .asCommutativeRing, and ensure it is correct.

non:
This commit makes sure we run our law tests on the ring
produced by Bool[Boolean].asCommutativeRing.

We might want to audit other typeclass-producing methods
to ensure that we do at least some law-checking on those
instances as well.

Review by @TomasMikula and @johnynek.


non:
(I confirmed that this test caught the error before I changed the `.asCommutativeRing` implementation.)

TomasMikula:
Looks good to me.

PR #99, opened 2015-10-14
Remove platform-specific rint implementation.

### LINKED ISSUE #74: Remove platform specific rint()
As rint() is now in scalajs - https://github.com/scala-js/scala-js/pull/1822 - the platform specific implementations can be remove when scala-js 0.6.5 is released

non:
By upgrading scala.js to 0.6.5, we can just use
java.lang.Math.rint and let scalajs wire it up
to a correct implementation.

This also allows the std module to become a
"pure" cross project, removing the need for
js/src, jvm/src, and shared/src.

Fixes #74. Review by @InTheNow.


johnynek:
Looks good to me, but I don't really know anything about how scala-js compilation works, so it is tough to see the impact of these changes.

InTheNow:
LGTM :+1:

non:
Thanks!

Issue #98, opened 2015-10-14
Add BoolRing type class


TomasMikula:
This is a proposal to add type class for [_Boolean rings_](https://en.wikipedia.org/wiki/Boolean_ring).

A Boolean ring is a ring that satisfies `x times x = x` for all `x` (i.e. every element is idempotent with respect to multiplication).

There isn't general consensus whether a Boolean ring assumes `1` (multiplicative identity). Some sources refer to Boolean ring as a Boolean algebra without identity. However, since `algebra.ring.Ring` has `1`, it makes sense for `BoolRing` to include `1` as well. Adding a Boolean ring without `1` could be addressed as a separate issue, possibly calling it `BoolRng`.

The above property implies:
 1. Every Boolean ring is a commutative ring.
 2. Every element is its own additive inverse, i.e. `x plus x = 0`.

As a consequence of 1., `BoolRing` should extend `algebra.ring.CommutativeRing`.

As a consequence of 2., `BoolRing` can implement `negate` to be

```scala
def negate(x: A): A = x
```

A Boolean ring is equivalent to a Boolean algebra, with [conversions both ways](https://en.wikipedia.org/wiki/Boolean_ring#Relation_to_Boolean_algebras).

`Bool#asCommutativeRing` already returns a Boolean ring. However, this fact is currently not captured by the return type.

johnynek:
Should we also have a method `.asBooleanAlgebra` or should this also subclass `Bool`?

non:
I slightly prefer `.asBooleanAlgebra`, since it avoids possible ambiguities between a `Bool[A]` and its corresponding `BoolRing[A]` (and also so that `Bool[A].asCommutativeRing.asBooleanAlgebra eq Bool[A]`.

But if you all prefer subclassing I think that would be OK here as well.

TomasMikula:
For a subclassing solution, it would be sufficient for `Bool` to extend `CommutativeRing`&mdash;no `BoolRing` needed. That was my original suggestion. Then @non pointed out that it could conflict with other, more natural implicit rings for the same type.

I will also note that if/when we add the versions without identity (i.e. a Boolean ring without identity, equivalent to a Boolean algebra without identity and relative complement instead of (absolute) complement), whether they are going to be the same type class or two different type classes will have to copy this decision.

I'm fine with either solution.

johnynek:
[...]

PR #102, opened 2015-10-28
Let Ring and CommutativeRing companions extend RingFunctions.

TomasMikula:
It seemed like an oversight that `object Ring` did not extend `RingFunctions`.

### DIFF
--- a/core/src/main/scala/algebra/ring/Ring.scala
+++ b/core/src/main/scala/algebra/ring/Ring.scala
@@ -32,11 +32,11 @@ trait Ring[@sp(Int, Long, Float, Double) A] extends Any with Rig[A] with Rng[A] 
   def fromInt(n: Int): A = sumN(one, n)
 }
 
-trait RingFunctions extends AdditiveGroupFunctions with MultiplicativeMonoidFunctions {
-  def fromInt[@sp(Int, Long, Float, Double) A](n: Int)(implicit ev: Ring[A]): A =
+trait RingFunctions[R[T] <: Ring[T]] extends AdditiveGroupFunctions with MultiplicativeMonoidFunctions {
+  def fromInt[@sp(Int, Long, Float, Double) A](n: Int)(implicit ev: R[A]): A =
     ev.fromInt(n)
 }
 
-object Ring extends RingFunctions {
+object Ring extends RingFunctions[Ring] {
   @inline final def apply[A](implicit ev: Ring[A]): Ring[A] = ev
 }
--- a/core/src/main/scala/algebra/ring/Ring.scala
+++ b/core/src/main/scala/algebra/ring/Ring.scala
@@ -32,11 +32,11 @@ trait Ring[@sp(Int, Long, Float, Double) A] extends Any with Rig[A] with Rng[A] 
   def fromInt(n: Int): A = sumN(one, n)
 }
 
-trait RingFunctions {
+trait RingFunctions extends AdditiveGroupFunctions with MultiplicativeMonoidFunctions {
   def fromInt[@sp(Int, Long, Float, Double) A](n: Int)(implicit ev: Ring[A]): A =
     ev.fromInt(n)
 }
 
-object Ring extends AdditiveGroupFunctions with MultiplicativeMonoidFunctions {
+object Ring extends RingFunctions {
   @inline final def apply[A](implicit ev: Ring[A]): Ring[A] = ev
 }
--- a/core/src/main/scala/algebra/ring/BoolRing.scala
+++ b/core/src/main/scala/algebra/ring/BoolRing.scala
@@ -35,6 +35,6 @@ private[ring] class BoolFromBoolRing[A](orig: BoolRing[A]) extends GenBoolFromBo
   override def asBoolRing: BoolRing[A] = orig
 }
 
-object BoolRing extends RingFunctions {
+object BoolRing extends RingFunctions[BoolRing] {
   @inline final def apply[A](implicit r: BoolRing[A]): BoolRing[A] = r
 }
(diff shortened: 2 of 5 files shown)


tixxit:
This LGTM, but should we also use this in EuclideanRingFunctions and FieldFunctions too?

TomasMikula:
Right, I will add those. I did this as a preparatory refactoring for #98, so I only did the minimum relevant for that and did not look around.

TomasMikula:
Added.

non:
Looks good. :+1:

core/src/main/scala/algebra/ring/Ring.scala
```diff
@@ -32,11 +32,11 @@ trait Ring[@sp(Int, Long, Float, Double) A] extends Any with Rig[A] with Rng[A]
   def fromInt(n: Int): A = sumN(one, n)
 }
 
-trait RingFunctions {
+trait RingFunctions extends AdditiveGroupFunctions with MultiplicativeMonoidFunctions {
```
johnynek: I wonder if these should be parameterized
```scala
trait RingFunctions[R[T] <: Ring[T]] {
  def fromInt[A](n: Int)(implicit ev: R[A]): A = ...
}
```
I'm worried there is some corner case of implicit resolution such that: `CommutativeRing[T].plus(a, b)` and `CommutativeRing.plus(a, b)` resolve to different implicits (one finds a `CommutativeRing[Int]` the other finds a `Ring[Int]`, which are not the same).

core/src/main/scala/algebra/ring/Ring.scala
```diff
@@ -32,11 +32,11 @@ trait Ring[@sp(Int, Long, Float, Double) A] extends Any with Rig[A] with Rng[A]
...
   def fromInt[@sp(Int, Long, Float, Double) A](n: Int)(implicit ev: Ring[A]): A =
     ev.fromInt(n)
 }
 
-object Ring extends AdditiveGroupFunctions with MultiplicativeMonoidFunctions {
+object Ring extends RingFunctions {
   @inline final def apply[A](implicit ev: Ring[A]): Ring[A] = ev
```
johnynek: I guess this could be a trait too:
```scala
trait Implicitly[R[_]] {
  @inline final def apply[A](implicit r: R[A]): R[A] = r
}
```

core/src/main/scala/algebra/ring/Ring.scala
non: I think if we had `CommutativeRing[A]` and `Ring[A]` at the same implicit precedence, it would create an ambiguity (and not compile). For this reason, I think that we usually work hard to ensure this isn't possible.

That said, parameterizing the trait on the type class type is not a bad idea -- but we would want to make sure we're doing it everywhere that is appropriate, not just here.

core/src/main/scala/algebra/ring/Ring.scala
non: This is one of those cases where for performance reasons I think using a shared trait is a bad idea -- once you use `Ring.apply`, `CommutativeRing.apply`, and `EuclideanRing.apply` all inlining would be de-optimized (due to the megamorphic call site).

In the long run I would prefer to use *imp* to compile this entire call away -- in that case using a trait would be fine. But until then I'd rather not make this change.

core/src/main/scala/algebra/ring/Ring.scala
tixxit: Scala should prioritize the most "specific" trait. So, if both a Ring[A] and a Field[A] are in scope, and we ask for an implicit Ring[A], then the field *should* be chosen. However, relying on this rule is pretty hairy (IMO) and I think we should parameterize.

TomasMikula: Hmm, wondering whether the `fromInt` method should be in `RingFunctions` at all. One can already write

```scala
Ring[A].fromInt(n)
```

with `fromInt` in `RingFunctions` (and thus in `Ring`), one can write

```scala
Ring.fromInt[A](n)
```

I don't see any benefit here.

johnynek: in the later case you wan write `Ring.fromInt(x)` and use type inference. In the former you can't.

But yeah, there is an argument to be made that we should just live with the version where we write the type explicitly.

TomasMikula: Ah, I see.

core/src/main/scala/algebra/ring/Ring.scala
johnynek: What is "this change"? The `Implicitly` trait, the type parameter in `RingFunctions` or this PR?

non: I meant using a shared `Implicitly` trait to inherit `apply`.

johnynek: So, you'd still be in support of putting the type parameter on `RingFunctions` so we can be more precise about the implicit resolved in, for instance, `CommutativeRing.plus(a, b)`?

non: Oh yes, sorry. I do think that's a good idea.

non:
Where does this PR stand? Did we decide to parameterize the `RingFunctions` trait?

johnynek:
I thought we did decide to parameterize RingFunctions.

non:
OK great, so we're waiting on that. Thanks!

TomasMikula:
Check out the new commit.

TomasMikula:
Oops, turns out I didn't merge the current master to my working copy.

TomasMikula:
Fixed, let's see what Travis says now.

non:
:+1:

PR #103, opened 2015-10-28
Make sure commutativity of CommutativeRing is tested.

TomasMikula:
It seems that commutativity of CommutativeRings (and EuclideanRings) was not tested.


non:
:+1: Thanks for catching this!

PR #104, opened 2015-10-28
Add BoolRing type class. Resolves #98.

### LINKED ISSUE #98: Add BoolRing type class

TomasMikula:


laws/shared/src/test/scala/algebra/laws/LawTests.scala
```diff
@@ -33,16 +33,23 @@ trait LawTestsBase extends FunSuite with Discipline {
...
+    LawChecker("[" + tag.tpe.toString + (if(extraTag != "") "@@" + extraTag else "") + "]", laws)
 
   laws[OrderLaws, Boolean].check(_.order)
   laws[LogicLaws, Boolean].check(_.bool)
   laws[LogicLaws, SimpleHeyting].check(_.heyting)
   laws[LatticePartialOrderLaws, Boolean].check(_.boundedLatticePartialOrder)
+  laws[RingLaws, Boolean].check(_.boolRing(BooleanRing))
+
+  // ensure that Bool[A].asCommutativeRing is a valid BoolRing
+  laws[RingLaws, Boolean]("ring-from-bool").check(_.boolRing(Bool[Boolean].asCommutativeRing))
```
TomasMikula: Since we now have two `RuleSet`s for `[RingLaws, Boolean]`, it was necessary to somehow distinguish the property names. This is done by an extra String tag `ring-from-bool`.

Similarly below for `[LogicLaws, Boolean]`.

core/src/main/scala/algebra/ring/BoolRing.scala
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * A Boolean ring is a ring whose multiplication is idempotent, that is
+ * `a⋅a = a` for all elements ''a''. This property also implies `a+a = 0`
+ * for all ''a'', and `a⋅b = b⋅a` (commutativity of multiplication).
+ */
+trait BoolRing[A] extends Any with CommutativeRing[A] { self =>
+  override final def negate(x: A): A = x
+
+  def asBool: Bool[A] = new Bool[A] {
```
johnynek: might be nice to add the documentation to just mentio what the mapping is here (and == times, for instance).

std/src/main/scala/algebra/std/boolean.scala
```diff
@@ -9,6 +10,13 @@ package object boolean extends BooleanInstances
 trait BooleanInstances {
   implicit val booleanAlgebra: BooleanAlgebra =
     new BooleanAlgebra
+
+  val BooleanRing = new BoolRing[Boolean] {
```
johnynek: this is actually a subset of the `Fields[Boolean]`. Is there any relationship as to when a `Field[T]` is an extension of a `BoolRing[T]`?

core/src/main/scala/algebra/lattice/Bool.scala
```diff
@@ -28,17 +28,17 @@ trait Bool[@sp(Int, Long) A] extends Any with Heyting[A] { self =>
...
   /**
-   * Every Boolean algebra is a CommutativeRing, but we don't extend
-   * CommutativeRing because, e.g. we might want a Bool[Int] and CommutativeRing[Int] to
+   * Every Boolean algebra is a BoolRing, but we don't extend BoolRing
+   * because, e.g. we might want a Bool[Int] and CommutativeRing[Int] to
    * refer to different structures, by default.
    */
-  def asCommutativeRing: CommutativeRing[A] =
-    new CommutativeRing[A] {
+  def asCommutativeRing: BoolRing[A] =
```
johnynek: should we also `overrride def asCommutativeRig: BoolRing[A] = asCommutativeRing`?

std/src/main/scala/algebra/std/boolean.scala
TomasMikula: Hmm, so `BoolRing` requires `a+a = 0`, thus specifically `1+1 = 0`, so only fields of characteristic 2 are boolean rings.

On the other side, `Field` requires a multiplicative inverse for every element except 0, i.e. `∀a≠0 ∃b a⋅b = 1`, which rewritten to lattice notation is `∀a≠0 ∃b a∧b = 1`, but this is only true for `a=1`. Thus any boolean ring with more than two elements will not be a field.

If the same operations are already implemented, I can remove this instance and make the `Field` instance extend `BoolRing`.

core/src/main/scala/algebra/ring/BoolRing.scala
TomasMikula: Will do.

core/src/main/scala/algebra/lattice/Bool.scala
TomasMikula: Well, that would be incompatible with the inherited `BoundedDistributiveLattice#asCommutativeRig`, which defines `plus` as `or` instead of `xor`

johnynek: Should we change the name perhaps to `asBoolRing`? It is easy to mistype `Ring` when instead of `Rig` or vice-versa.

It's a bit weird that if you just need a `Rig` there are two ways to get it: `asCommutativeRig` and `asCommutativeRing` on a Bool. We added the comment in the README.md about this. Maybe that documentation should also appear here.

TomasMikula: +1 for renaming. Should I keep and deprecated `asCommutativeRing`, or just delete it?

I will add a comment that it is not an extension of `asCommutativeRig`.

std/src/main/scala/algebra/std/boolean.scala
TomasMikula: @johnynek I can't find an instance of `Field[Boolean]` in `algebra`. Am I just not looking in the right places?

core/src/main/scala/algebra/lattice/Bool.scala
johnynek: I would delete it. The next version of the library (0.4.0) is already going to be binary incompatible with 0.3.1 and I assume the impact of this method rename is small. Our goal at this stage is to get the core stabilized.

std/src/main/scala/algebra/std/boolean.scala
johnynek: I guess we didn't implement it. Algebird has one.

TomasMikula: Should I rename it to `GF2` and make it implement `Field[Boolean]`?

johnynek: up to you. If so, we could do `Field[Boolean] with BoolRing[Boolean]`. I'm not sure I'd bother in this PR, personally.

johnynek:
:+1: if @non doesn't have any comments, I say merge.

non:
This looks good to me. I'm fine with adding GF2 in a future PR. :+1:

Issue #108, opened 2015-10-31
Should we remove the construction methods in algebra.lattice?


johnynek:
In the lattices, we have a number of methods to create ring structures from lattices. These seem like particular constructions to me, and perhaps not best bolted onto the traits themselves.

For instance, we could add:
```scala
trait Semigroup[T] {
  //...
  def asMonoid: Monoid[Option[T]]
}
```
which is a construction to create a Monoid from a semigroup using `None` as the unit, but I wouldn't want to do that. Similarly we can build a `Ring[T]` from a `Rng[T]` by adding a `1` element, but again, that construction I would rather have a named class for that than bolt it onto the trait.

By contrast, `.meetSemilattice` is not a construction, it is exactly taking the `meet` and returning a `Semilattice` instance which `combine == meet`. This feels different to me.

Somewhat similarly is the `.dual`. Maybe we should have `case class DualLattice[T](lattice: Lattice[A]) extends Lattice[A]` or some such, instead of methods on the traits.

TomasMikula:
Note that the return type of `dual` is overridden in more specific lattices, so one `DualLattice` class will not cover them all.

johnynek:
Yeah. `dual` we can either set aside from this discussion, or make `DualX` for each X. I can't see another way to do that.

non:
I would be fine removing these things. It may be the case that we will want to use something more like `Dual[Bool, A]` to do something like this.

non:
Given all the confusion around these methods (what to call them, what their return types should be, whether they will be refined in subtypes, etc.) I think we should just remove them and handle this via a different mechanism.

TomasMikula:
Yeah, it makes more and more sense to me :+1:

PR #109, opened 2015-10-31
Add BoolRng and GenBool. Resolves #107.

### LINKED ISSUE #107: Type classes for Boolean ring and Boolean algebra without 1.
There are useful data types that are almost like `Bool` or `BoolRing`, except they don't have a top element. A good example is `Set[A]`&mdash;when `A` is a type with infinitely many inhabitants, we cannot have the top element (`Set` of all `A`s), but it is still useful to view it as a Boolean algebra.

#### Naming

A suitable name for a `BoolRing` without identity seems to be **BoolRng**.

For `Bool` without top, the best I have so far is **ToplessBool**. I'm not completely satisfied with this name, because it might suggest `Bool` _necessarily_ without top, while what we mean is `Bool` _potentially_ without top.

#### Definitions

`BoolRng` is just a `Rng` with idempotent multiplication (`a⋅a=a`).

`ToplessBool` is a `DistributiveLattice` with bottom and for all elements _a, b_, there is an element `b\a`, called the _relative complement of a in b_, such that

1. a ∧ b\a = 0
2. a ∨ b\a = a ∨ b

Note that in the presence of 1, we can define `¬a = 1\a` and the above laws give the familiar laws of Boolean algebra:

a ∧ ¬a = a ∧ 1\a = 0
a ∨ ¬a = a ∨ 1\a = a ∨ 1 = 1

#### Equivalence

`BoolRng` and `ToplessBool` are equivalent.

Given a `ToplessBool`, one can construct a `BoolRng` in the following way:

 - 0 = 0
 - a⋅b = a∧b
 - a+b = a\b ∨ b\a
 - -a = a

Given a `BoolRng`, one can construct a `ToplessBool` in the following way:

 - 0 = 0
 - a∧b = a⋅b
 - a∨b = a+b+(a⋅b)
 - a\b = (a+b)⋅a = a + a⋅b

TomasMikula:
This is a work in progress on topless Bool/BoolRing (#107), subject to some renaming. In this version I used the names `GenBool` for _generalized Boolean algebra_ and `without(a, b)` for `a\b` (relative complement of b with respect to a).

### DIFF
--- a/core/src/main/scala/algebra/lattice/GenBool.scala
+++ b/core/src/main/scala/algebra/lattice/GenBool.scala
@@ -1,0 +1,57 @@
+package algebra
+package lattice
+
+import ring.BoolRng
+import scala.{specialized => sp}
+
+/**
+ * Generalized Boolean algebra, that is, a Boolean algebra without
+ * the top element. Generalized Boolean algebras do not (in general)
+ * have (absolute) complements, but they have ''relative complements''
+ * (see [[GenBool.without]]).
+ */
+trait GenBool[@sp(Int, Long) A] extends Any with DistributiveLattice[A] with BoundedJoinSemilattice[A] { self =>
+  def and(a: A, b: A): A
+  override def meet(a: A, b: A): A = and(a, b)
+
+  def or(a: A, b: A): A
+  override def join(a: A, b: A): A = or(a, b)
+
+  /**
+   * The operation of ''relative complement'', symbolically often denoted
+   * `a\b` (the symbol for set-theoretic difference, which is the
+   * meaning of relative complement in the lattice of sets).
+   */
+  def without(a: A, b: A): A
+
+  /**
+   * Logical exclusive or, set-theoretic symmetric difference.
+   * Defined as `a\b ∨ b\a`.
+   */
+  def xor(a: A, b: A): A = or(without(a, b), without(b, a))
+
+  /**
+   * Every generalized Boolean algebra is also a `BoolRng`, with
+   * multiplication defined as `and` and addition defined as `xor`.
+   */
+  def asBoolRing: BoolRng[A] = new BoolRngFromGenBool(self)
+}
+
+private[lattice] class BoolRngFromGenBool[@sp(Int, Long) A](orig: GenBool[A]) extends BoolRng[A] {
+  def zero: A = orig.zero
+  def plus(x: A, y: A): A = orig.xor(x, y)
+  def times(x: A, y: A): A = orig.and(x, y)
+  override def asBool: GenBool[A] = orig
+}
+
+trait GenBoolFunctions {
+  def zero[@sp(Int, Long) A](implicit ev: GenBool[A]): A = ev.zero
+  def and[@sp(Int, Long) A](x: A, y: A)(implicit ev: GenBool[A]): A = ev.and(x, y)
+  def or[@sp(Int, Long) A](x: A, y: A)(implicit ev: GenBool[A]): A = ev.or(x, y)
+  def without[@sp(Int, Long) A](x: A, y: A)(implicit ev: GenBool[A]): A = ev.without(x, y)
+  def xor[@sp(Int, Long) A](x: A, y: A)(implicit ev: GenBool[A]): A = ev.xor(x, y)
+}
+
+object GenBool extends BoolFunctions {
+  @inline final def apply[@sp(Int, Long) A](implicit ev: GenBool[A]): GenBool[A] = ev
+}
--- a/laws/shared/src/main/scala/algebra/laws/LogicLaws.scala
+++ b/laws/shared/src/main/scala/algebra/laws/LogicLaws.scala
@@ -1,7 +1,7 @@
 package algebra.laws
 
 import algebra._
-import algebra.lattice.{Heyting, Bool}
+import algebra.lattice.{Heyting, Bool, GenBool}
 
 import org.typelevel.discipline.Laws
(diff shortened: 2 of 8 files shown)


laws/shared/src/test/scala/algebra/laws/LawTests.scala
```diff
@@ -68,8 +68,10 @@ trait LawTestsBase extends FunSuite with Discipline {
   laws[OrderLaws, List[String]].check(_.order)
   laws[GroupLaws, List[String]].check(_.monoid)
 
-  laws[LatticeLaws, Set[Int]].check(_.distributiveLattice)
-  laws[LatticeLaws, Set[Int]].check(_.boundedJoinLattice)
+  laws[LogicLaws, Set[Byte]].check(_.generalizedBool)
+  laws[RingLaws, Set[Byte]].check(_.boolRng(setBoolRng[Byte]))
+  laws[LogicLaws, Set[Byte]]("bool-from-rng").check(_.generalizedBool(setBoolRng.asBool))
+  laws[RingLaws, Set[Byte]]("rng-from-bool").check(_.boolRng(GenBool[Set[Byte]].asBoolRing))
```
TomasMikula: Switched from `Set[Int]` to `Set[Byte]` in order to have more exhaustive coverage of the domain.

laws/shared/src/main/scala/algebra/laws/LogicLaws.scala
```diff
@@ -56,9 +56,26 @@ trait LogicLaws[A] extends LatticeLaws[A] {
...
+    name = "generalized bool",
+    parents = Seq(),
+    ll = new LatticeProperties(
+      name = "lowerBoundedDistributiveLattice",
+      parents = Seq(boundedJoinSemilattice, distributiveLattice),
+      join = Some(boundedSemilattice(A.joinSemilattice)),
+      meet = Some(semilattice(A.meetSemilattice))
+    ),
+
+    "x∖y ∧ y = 0" -> forAll { (x: A, y: A) =>
```
johnynek: Did you want "y\x" for "x y" here and below?

core/src/main/scala/algebra/lattice/GenBool.scala
```diff
@@ -0,0 +1,57 @@
...
+   * Logical exclusive or, set-theoretic symmetric difference.
+   * Defined as `a\b ∨ b\a`.
+   */
+  def xor(a: A, b: A): A = or(without(a, b), without(b, a))
+
+  /**
+   * Every generalized Boolean algebra is also a `BoolRng`, with
+   * multiplication defined as `and` and addition defined as `xor`.
+   */
+  def asBoolRing: BoolRng[A] = new BoolRngFromGenBool(self)
```
johnynek: `asBoolRng`, right?

laws/shared/src/main/scala/algebra/laws/LogicLaws.scala
TomasMikula: Hmm, there is a Unicode set difference symbol instead of a regular backslash used here because it is inside a string literal. Turns out they are not rendered correctly :/

core/src/main/scala/algebra/lattice/GenBool.scala
TomasMikula: This was intentional, because I didn't want to introduce a new method name.
It is kind of analogous to dual, where the subtype overrides the return
type.

johnynek: Should we change both method names to `BoolRng`? It seems a bit strange to name it for the more derived case.

TomasMikula: That would make sense. Or maybe to asRingLike? Similarly asBool to
asGenBool/asBoolLike?

johnynek: Should we settle #108 first? Because, my first preference is actually to remove these methods and make the classes you wrote public.

TomasMikula: Yeah, maybe.

TomasMikula:
Although I used the name `GenBool` for _generalized Boolean algebra_, I came to like the shortcut less and less, because  "Gen" is kind of ambiguous (generic?, `org.scalacheck.Gen`-related?). I guess I would prefer spelling out `GeneralizedBool`.

core/src/main/scala/algebra/lattice/GenBool.scala
johnynek: Actually, this PR is basically done. Let's merge this and not block it on #108. If we remove constructions, this is a marginal amount compared with what we have already.

non:
I'm fine with either of those names. I like @johnynek's idea of just merging this as-is but if you'd rather I would be fine with changing the name.

TomasMikula:
If you are fine with it, then go ahead and merge. How do people feel about `without` as the name for _relative complement_. It is an intuitive name for the case of set difference. For logic, the intuition is `a butNot b` (as a primitive operation, because we don't have `not`/`complement`).

non:
:+1:

non:
So the idea is that we'd write *the relative complement of a in b*  as `a \ b` and `without(b, a)`?

TomasMikula:
Yes, exactly, so that the argument order of the spelled out name is consistent with the symbolic notation.

non:
Sounds good to me.

johnynek:
I think without is fine.

PR #111, opened 2015-11-02
Simplify RingLaws: Don't pass around a Predicate, be more explicit about requiring non-zero elements.

TomasMikula:
Hi,

I present this PR for discussion. The goal is to simplify RingLaws, namely the case when we need a generator of non-zero values. The approach is to be more explicit about what laws require non-zero elements. I added an extra `Arbitrary[A]` argument to `group` properties and "descendants", which then allow to pass a generator of non-zero values from `multiplicativeGroup` and `multiplicativeCommutativeGroup`. `multiplicativeGroup` and `multiplicativeCommutativeGroup` now also explicitly require an instance of `AdditiveMonoid` in order to access `zero` to create non-zero generator.

I find it slightly clearer than the predicate approach. Whether you agree is up for discussion, but by the sheer number of lines saved, it is a simplification.

### DIFF
--- a/laws/shared/src/main/scala/algebra/laws/RingLaws.scala
+++ b/laws/shared/src/main/scala/algebra/laws/RingLaws.scala
@@ -3,37 +3,25 @@ package laws
 
 import algebra.ring._
 
-import org.typelevel.discipline.{Laws, Predicate}
+import org.typelevel.discipline.Laws
 
 import org.scalacheck.{Arbitrary, Prop}
 import org.scalacheck.Arbitrary._
 import org.scalacheck.Prop._
 
 object RingLaws {
-  def apply[A : Eq : Arbitrary](implicit pred0: Predicate[A]) = new RingLaws[A] {
+  def apply[A : Eq : Arbitrary] = new RingLaws[A] {
     def Arb = implicitly[Arbitrary[A]]
-    def pred = pred0
-    val nonZeroLaws = new GroupLaws[A] {
-      def Arb = Arbitrary(arbitrary[A] filter pred0)
-      def Equ = Eq[A]
-    }
+    def Equ: Eq[A] = Eq[A]
   }
 }
 
 trait RingLaws[A] extends GroupLaws[A] {
 
-  // must be a val (stable identifier)
-  val nonZeroLaws: GroupLaws[A]
-  def pred: Predicate[A]
-
-  def withPred(pred0: Predicate[A], replace: Boolean = true): RingLaws[A] = RingLaws[A](
-    Equ,
-    Arb,
-    if (replace) pred0 else pred && pred0
-  )
-
   implicit def Arb: Arbitrary[A]
-  implicit def Equ: Eq[A] = nonZeroLaws.Equ
+  implicit def Equ: Eq[A]
+
+  private def nonZeroArb(implicit am: AdditiveMonoid[A]): Arbitrary[A] = Arbitrary(arbitrary[A] filter { a: A => Equ.neqv(a, am.zero) })
 
   // multiplicative groups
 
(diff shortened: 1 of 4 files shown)


non:
Hi Tomas,

Sorry I didn't see this earlier. I need to have a closer look to how things are threaded through. I have two (possibly unfounded) concerns:

1. I worry that we will accidentally use the *non-zero arbitrary* implicitly in a place where we want the *full arbitrary* and end up failing to test certain properties with zero.

2. In general we have tried to move away from using `.filter` on `Arbitrary[_]` instances to avoid cases where ScalaCheck aborts due to throwing out too many test cases. Removing one element may be OK, but I think some generators end up creating tons of zeros. The usual fix is to use `.map` and replace zero with a non-zero value. In this generic context we could use one (since I think all the "non-zero" properties are involving rings).

laws/shared/src/main/scala/algebra/laws/RingLaws.scala
```diff
@@ -3,37 +3,25 @@ package laws
...
 import org.scalacheck.Arbitrary._
 import org.scalacheck.Prop._
 
 object RingLaws {
-  def apply[A : Eq : Arbitrary](implicit pred0: Predicate[A]) = new RingLaws[A] {
+  def apply[A : Eq : Arbitrary] = new RingLaws[A] {
     def Arb = implicitly[Arbitrary[A]]
-    def pred = pred0
-    val nonZeroLaws = new GroupLaws[A] {
-      def Arb = Arbitrary(arbitrary[A] filter pred0)
```
TomasMikula: @non Ad your point 2., I agree we should avoid `filter`, but I didn't introduce a new `filter`, I just moved the one above to a different location.

laws/shared/src/main/scala/algebra/laws/RingLaws.scala
```diff
@@ -3,37 +3,25 @@ package laws
...
-    Equ,
-    Arb,
-    if (replace) pred0 else pred && pred0
-  )
-
   implicit def Arb: Arbitrary[A]
-  implicit def Equ: Eq[A] = nonZeroLaws.Equ
+  implicit def Equ: Eq[A]
+
+  private def nonZeroArb(implicit am: AdditiveMonoid[A]): Arbitrary[A] = Arbitrary(arbitrary[A] filter { a: A => Equ.neqv(a, am.zero) })
```
TomasMikula: @non Ad your point 1., notice that `nonZeroArb` is private and not implicit.

laws/shared/src/main/scala/algebra/laws/RingLaws.scala
```diff
@@ -57,17 +45,19 @@ trait RingLaws[A] extends GroupLaws[A] {
...
-  )
+  def multiplicativeGroup(implicit A: MultiplicativeGroup[A] with AdditiveMonoid[A]) = {
+    implicit val Arb: Arbitrary[A] = nonZeroArb(A)
+    new MultiplicativeProperties(
+      base = _.group(A.multiplicative, Arb),
+      parent = Some(multiplicativeMonoid),
+      "consistent division" -> forAll{ (x: A, y: A) =>
+        (A.div(x, y) ?== A.times(x, A.reciprocal(y)))
+      }
+    )
```
TomasMikula: The non-zero arbitrary `Arb` is implicit here in the local scope only. The only reason it is there implicitly is to use it in the `"consistent division"` property. If you can think of a clear way to pass it to that property explicitly, or to limit the scope of the implicit value even more, I will make a change.

johnynek: I think:
```scala
val gen = nonZeroArb(A).arbitrary
forAll(gen, gen) { ...
```
Will do it.

TomasMikula:
I removed the implicit instance of non-zero arbitrary. (Thanks @johnynek for the tip.)

johnynek:
seems stale, so closing. Happy to revisit.

PR #115, opened 2015-11-12
Add Order.andThen and Monoid[Order[A]]

ceedubs:
Hopefully the comments in the code provide sufficient explanation.

This is something I use on scalaz's `Order` very often, and to be honest I'm not sure what people do without it on the standard library's `Ordering`. What do people do when they want to sort a list, comparing items first by one field then by another?

Sorry, the law-testing is pretty gross. Please do let me know if you have a better alternative, and I'm happy to make a change.

### DIFF
--- a/core/src/main/scala/algebra/Order.scala
+++ b/core/src/main/scala/algebra/Order.scala
@@ -95,6 +95,21 @@ trait Order[@sp A] extends Any with PartialOrder[A] { self =>
    */
   override def gt(x: A, y: A): Boolean =
     compare(x, y) > 0
+
+  /**
+   * Returns a new `Order[A]` instance that first compares by the original
+   * `Order` instance and uses the provided `Order` instance to "break ties".
+   *
+   * That is, `x.combine(y)` creates an `Order` that first orders by `x` and
+   * then (if two elements are equal) falls back to `y` for the comparison.
+   */
+  def andThen(o: Order[A]): Order[A] = new Order[A] {
+    def compare(x: A, y: A) = {
+      val c = self.compare(x, y)
+      if (c == 0) o.compare(x, y)
+      else c
+    }
+  }
 }
 
 trait OrderFunctions {
@@ -148,6 +163,27 @@ object Order extends OrderFunctions {
    */
   implicit def ordering[A](implicit ev: Order[A]): Ordering[A] =
     new Ordering[A] {
-      def compare(x: A, y: A) = ev.compare(x, y)
+      def compare(x: A, y: A): Int = ev.compare(x, y)
+    }
+
+  /**
+   * A `Monoid[Order[A]]` can be generated for all `A` with the following
+   * properties:
+   *
+   * `empty` returns a trivial `Order[A]` which considers all `A` instances to
+   * be equal.
+   *
+   * `combine(x: Order[A], y: Order[A])` creates an `Order[A]` that first
+   * orders by `x` and then (if two elements are equal) falls back to `y`.
+   *
+   * @see [[Order.andThen]]
+   */
+  implicit def orderMonoid[A]: Monoid[Order[A]] =
+    new Monoid[Order[A]] {
+      val empty: Order[A] = new Order[A] {
+        def compare(x: A, y: A): Int = 0
+      }
+
+      def combine(x: Order[A], y: Order[A]): Order[A] = x andThen y
     }
 }
--- a/laws/shared/src/test/scala/algebra/laws/LawTests.scala
+++ b/laws/shared/src/test/scala/algebra/laws/LawTests.scala
@@ -9,7 +9,7 @@ import algebra.std.Rat
 
 import org.typelevel.discipline.{Laws, Predicate}
 import org.typelevel.discipline.scalatest.Discipline
-import org.scalacheck.Arbitrary
+import org.scalacheck.{Arbitrary, Gen}, Arbitrary.arbitrary
 import org.scalatest.FunSuite
 
 trait LawTestsBase extends FunSuite with Discipline {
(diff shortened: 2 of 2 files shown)


core/src/main/scala/algebra/Order.scala
```diff
@@ -95,6 +95,21 @@ trait Order[@sp A] extends Any with PartialOrder[A] { self =>
    */
   override def gt(x: A, y: A): Boolean =
     compare(x, y) > 0
+
+  /**
+   * Returns a new `Order[A]` instance that first compares by the original
+   * `Order` instance and uses the provided `Order` instance to "break ties".
+   *
+   * That is, `x.combine(y)` creates an `Order` that first orders by `x` and
```
johnynek: You call this `combine` here. I guess that is a copy/paste error.

core/src/main/scala/algebra/Order.scala
```diff
@@ -95,6 +95,21 @@ trait Order[@sp A] extends Any with PartialOrder[A] { self =>
...
+   * then (if two elements are equal) falls back to `y` for the comparison.
+   */
+  def andThen(o: Order[A]): Order[A] = new Order[A] {
```
johnynek: what about a more descriptive name. Like `whenEqual(o: Order[A])` or something.

laws/shared/src/test/scala/algebra/laws/LawTests.scala
```diff
@@ -143,5 +143,32 @@ trait LawTestsBase extends FunSuite with Discipline {
...
+
+  {
+    // This "arbitrary int order" isn't that arbitrary.
+    // I think this could be improved once we have a version of Scalacheck with Cogen.
+    implicit val arbOrderInt: Arbitrary[Order[Int]] = Arbitrary(
+      Gen.oneOf(
+        Gen.const(Order[Int]),
+        Gen.const(Order[Int].reverse)))
+
+    // This is a hack to fake an `Eq` instance for an `Order`.
```
johnynek: cute trick.

core/src/main/scala/algebra/Order.scala
```diff
@@ -148,6 +163,27 @@ object Order extends OrderFunctions {
...
+   * be equal.
+   *
+   * `combine(x: Order[A], y: Order[A])` creates an `Order[A]` that first
+   * orders by `x` and then (if two elements are equal) falls back to `y`.
+   *
+   * @see [[Order.andThen]]
+   */
+  implicit def orderMonoid[A]: Monoid[Order[A]] =
+    new Monoid[Order[A]] {
+      val empty: Order[A] = new Order[A] {
```
johnynek: I think this could be lifted out to `Order.allEqual` or something. This could be useful when people need an order, but don't care, and don't want to do the roundabout of `Monoid.empty[Order[A]]` which is strange thing to read when the rest of the code might have nothing to do with Monoids.

core/src/main/scala/algebra/Order.scala
```diff
@@ -148,6 +163,27 @@ object Order extends OrderFunctions {
...
+   *
+   * `empty` returns a trivial `Order[A]` which considers all `A` instances to
+   * be equal.
+   *
+   * `combine(x: Order[A], y: Order[A])` creates an `Order[A]` that first
+   * orders by `x` and then (if two elements are equal) falls back to `y`.
+   *
+   * @see [[Order.andThen]]
+   */
+  implicit def orderMonoid[A]: Monoid[Order[A]] =
```
johnynek: we have generally shied away from creating default instances and leave those to standard.

How attached are you to this being implicit? We generally don't make implicits available in core, but in std. Also, can we imagine another monoid? For instance, trivially the order of left and right could be swapped (y whenEqual x). Are there others? For instance what if you take the first equal as a result, and on non-equal go to the next one (is that associative?)

laws/shared/src/test/scala/algebra/laws/LawTests.scala
tixxit: As an alternative (though this is fine), you could also just literally fix the input type to only contain some small, fixed # of elements (instead of `Int`). Then we could just sort 2 lists using 2 different orders to check equality.

```scala
case class N100(n: Int) { require(n >= 0 && n < 100) }
implicit val arbN100 = Arbitrary(Gen.choose(0, 100).map(N100))
implicit val argN100Order = arbitrary[Int].map { seed =>
  val order = new Random(seed).shuffle(Vector.range(0, 100))
  Order.by { (n: N100) => order(n.n) }
}
implicit val N100OrderEq: Eq[Order[N100]] = Eq.by { implicit order => Vector.tabulate(100)(N100).sorted.map(_.n) }
```

core/src/main/scala/algebra/Order.scala
ceedubs: Oops good catch.

core/src/main/scala/algebra/Order.scala
ceedubs: Good idea.

core/src/main/scala/algebra/Order.scala
ceedubs: I don't think I'm too attached to this being implicit as long as there is a method for combining `Order` instances in this manner on the `Order` trait.

ceedubs:
Thanks for the feedback. I'll follow up with changes. Hopefully within the next few days.

core/src/main/scala/algebra/Order.scala
ceedubs: When I named it "andThen" I was thinking that when I talk about ordering I might say something like "order by X and then Y and then Z". But I'm not attached to this name. I'll change it to `whenEqual`.

laws/shared/src/test/scala/algebra/laws/LawTests.scala
ceedubs: Thanks @tixxit. I do think that comparing the sorting results is a better approach. Is the motivation for `N100` to have a smaller domain so that equal elements are more likely?

tixxit: It's really just about making the comparison feasible (something smaller would probably be even better). We'd still have too many permutations to expect equality to be common at all, but we're really just more concerned about the instances that aren't equal anyways, for the counter examples (eg `(ab)c != a(bc)`).

ceedubs:
I've updated my PR based on the provided feedback.

PR #119, opened 2015-11-15
Override reverse in Order.reverse value

ceedubs:
Override the `reverse` implementation in the result of `Order.reverse`
to simply return the original `Order` instance.


tixxit:
:+1:

PR #121, opened 2015-11-18
Add Order.fromOrdering method

ceedubs:
This provides a conversion from `scala.math.Ordering` to
`algebra.Order`.

I might be abusing the `extraTag` form of the `laws` function, I'm not
sure. If I don't provide a label then I get a duplicate test name error.


johnynek:
:+1: @non any objection?

TomasMikula:
@ceedubs Not abusing, this is what `extraTag` was introduced for. Otherwise it results in exactly the same property names as

```scala
laws[OrderLaws, Int].check(_.order)
```

in the same test suite.

:+1:

PR #122, opened 2015-11-20
Add Eq and PartialOrder instances for List.

### LINKED ISSUE #118: Eq and PartialOrder instances for List
Currently algebra provides an `Order[List[A]]` instance provided that `A` has an `Order` instance. However, it could also provide `Eq[List[A]]` with the weaker constraint that `A` has an `Eq` instance and similarly for `PartialOrder`.

Such [instances](https://github.com/non/cats/blob/692e85501eda19855fcd26e3f661ab9914961b64/core/src/main/scala/cats/std/list.scala#L81-L122) already exist in Cats. They should probably be moved to algebra.

non:
Fixes #118.

Note that if we want to support "pairwise instances" for additive/multiplicative semigroups for lists they would be really easy to do (treating uneven lengths as one/zero padded).

### DIFF
--- a/std/src/main/scala/algebra/std/list.scala
+++ b/std/src/main/scala/algebra/std/list.scala
@@ -6,12 +6,12 @@ import scala.collection.mutable
 
 package object list extends ListInstances 
 
-trait ListInstances {
+trait ListInstances extends ListInstances1 {
   implicit def listOrder[A: Order] = new ListOrder[A]
   implicit def listMonoid[A] = new ListMonoid[A]
 }
 
-trait ListInstances1 {
+trait ListInstances1 extends ListInstances2 {
   implicit def listPartialOrder[A: PartialOrder] = new ListPartialOrder[A]
 }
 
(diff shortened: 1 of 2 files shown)


std/src/main/scala/algebra/std/list.scala
```diff
@@ -11,6 +11,14 @@ trait ListInstances {
   implicit def listMonoid[A] = new ListMonoid[A]
 }
 
+trait ListInstances1 {
+  implicit def listPartialOrder[A: PartialOrder] = new ListPartialOrder[A]
+}
+
+trait ListInstances2 {
```
ceedubs: Are people meant to extend these traits directly? I would have expected `trait ListInstances1 extends ListInstances2` and `trait ListInstances extends ListInstances1` or something like that.

ceedubs:
Can we add law-checking for these?

std/src/main/scala/algebra/std/list.scala
non: Oh, haha, total oversight. I'll fix it. Thanks!

non:
Yes, I'll enable the additional law-checking too. Thanks for keeping me honest @ceedubs!

johnynek:
+1 for tests.

Could we factor the inner loop of these functions to only consume iterators without a perf cost? Then we could easily make instances for other `Iterable` types?

non:
@johnynek Yeah `Iterator` and `Builder` might work for everything except `Array`. I'll give it a shot.

PR #124, opened 2015-11-22
Add parameters to all functions that extend/are extended

### LINKED ISSUE #110: Copy the RingFunctions approach for others
`Additive*Functions` and `Multiplicative*Functions` could probably use the same treatment of having a type parameter.

johnynek:
fixes #110 

This uncovered one bug: `CommutativeRig` had `AdditiveGroupFunctions` but that would never have worked for any `Rig` because it is not a group.

Also on a personal note: I prepared and submitted this PR over the state of Colorado on a flight to New Jersey.

### DIFF
--- a/ring/src/main/scala/algebra/ring/CommutativeRig.scala
+++ b/ring/src/main/scala/algebra/ring/CommutativeRig.scala
@@ -8,6 +8,6 @@ import scala.{specialized => sp}
  */
 trait CommutativeRig[@sp(Int, Long, Float, Double) A] extends Any with Rig[A] with MultiplicativeCommutativeMonoid[A]
 
-object CommutativeRig extends AdditiveGroupFunctions with MultiplicativeMonoidFunctions {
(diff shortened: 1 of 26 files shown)


ring/src/main/scala/algebra/ring/CommutativeRig.scala
```diff
@@ -8,6 +8,6 @@ import scala.{specialized => sp}
  */
 trait CommutativeRig[@sp(Int, Long, Float, Double) A] extends Any with Rig[A] with MultiplicativeCommutativeMonoid[A]
 
-object CommutativeRig extends AdditiveGroupFunctions with MultiplicativeMonoidFunctions {
+object CommutativeRig extends AdditiveMonoidFunctions[CommutativeRig] with MultiplicativeMonoidFunctions[CommutativeRig] {
```
johnynek: here was the bug.

TomasMikula:
:+1:

non:
:+1:

PR #125, opened 2015-11-22
Add semilattices for Eq and Eq.and/or

johnynek:
These are similar to #115.

Mostly just for fun, I guess.

Along with #115 it seems that #70 may be redundant now (or at least can be implemented with the monoids/semilattices we have here):

```scala
def fromAll[A](fns: (A, A) => Boolean*): Eq[A] = 
  allEqualBoundedSemilattice.combineAll(fns.map(instance))
```


non:
:+1:

PR #126, opened 2015-11-22
Add GenBool instance for Map[K, V: GenBool].

TomasMikula:


std/src/main/scala/algebra/std/map.scala
```diff
@@ -130,3 +133,25 @@ class MapRng[K, V](implicit V: Rng[V]) extends MapSemiring[K, V] with Rng[Map[K,
   override def minus(x: Map[K, V], y: Map[K, V]): Map[K, V] =
     subtractMap(x, y)(V.minus)(V.negate)
 }
+
+class MapGenBool[K, V](implicit GBV: GenBool[V], EqV: Eq[V]) extends GenBool[Map[K, V]] {
+  def zero: Map[K, V] = Map.empty
+  
+  def and(a: Map[K, V], b: Map[K, V]): Map[K, V] = (
+    (a.keySet intersect b.keySet)
```
johnynek: for performance you might want:
```scala
(a.keySet intersect b.keySet).iterator
  .map { k => (k, GBV.and(a(k), b(k))) }
  .filter { case (_, v) => EqV.neqv(GBV.zero, v) }
  .toMap
```
putting the iterator early prevents materialization after .map and .filter. Also, note, .map is returning a set here and so if filter, so that is going to be bit slower than the iterator.

std/src/main/scala/algebra/std/map.scala
```diff
@@ -130,3 +133,25 @@ class MapRng[K, V](implicit V: Rng[V]) extends MapSemiring[K, V] with Rng[Map[K,
   override def minus(x: Map[K, V], y: Map[K, V]): Map[K, V] =
     subtractMap(x, y)(V.minus)(V.negate)
 }
+
+class MapGenBool[K, V](implicit GBV: GenBool[V], EqV: Eq[V]) extends GenBool[Map[K, V]] {
```
johnynek: this is sparse, unlike the other implementations above. We could take `Eq[V]` on anything with zero on the values, and do something similar. We did this in algebird (we messed up a bit and didn't define an Eq or Equiv, so we just use `==`), and often the sparseness was an issues.

I think in general, have a method to remove zeros is a better approach and less surprising (and also consistent with other items).

What do you think of the above?

std/src/main/scala/algebra/std/map.scala
```diff
@@ -130,3 +133,25 @@ class MapRng[K, V](implicit V: Rng[V]) extends MapSemiring[K, V] with Rng[Map[K,
...
+    filter { (kv: (K, V)) => EqV.neqv(GBV.zero, kv._2) }
+  ).toMap
+
+  def or(a: Map[K, V], b: Map[K, V]): Map[K, V] = (
+    (a.keySet union b.keySet)
+    map { k => (k, GBV.or(a.getOrElse(k, GBV.zero), b.getOrElse(k, GBV.zero))) }
+  ).toMap
+
+  def without(a: Map[K, V], b: Map[K, V]): Map[K, V] = (
+    a.iterator map { kv => b.get(kv._1) match {
```
johnynek: what about `map { case (k, v) => ` to make it clearer below?

std/src/main/scala/algebra/std/map.scala
```diff
@@ -130,3 +133,25 @@ class MapRng[K, V](implicit V: Rng[V]) extends MapSemiring[K, V] with Rng[Map[K,
...
+  
+  def and(a: Map[K, V], b: Map[K, V]): Map[K, V] = (
+    (a.keySet intersect b.keySet)
+    map { k => (k, GBV.and(a(k), b(k))) }
+    filter { (kv: (K, V)) => EqV.neqv(GBV.zero, kv._2) }
+  ).toMap
+
+  def or(a: Map[K, V], b: Map[K, V]): Map[K, V] = (
+    (a.keySet union b.keySet)
+    map { k => (k, GBV.or(a.getOrElse(k, GBV.zero), b.getOrElse(k, GBV.zero))) }
```
johnynek: like above, this is creating an intermediate set. Probably better to do `.iterator` after the union.

std/src/main/scala/algebra/std/map.scala
```diff
@@ -130,3 +133,25 @@ class MapRng[K, V](implicit V: Rng[V]) extends MapSemiring[K, V] with Rng[Map[K,
...
+  ).toMap
+
+  def without(a: Map[K, V], b: Map[K, V]): Map[K, V] = (
+    a.iterator map { kv => b.get(kv._1) match {
+      case Some(v2) => (kv._1, GBV.without(kv._2, v2))
+      case None => kv
+    }} filter { kv => EqV.neqv(GBV.zero, kv._2) }
```
johnynek: again, I'm not sure we always want this filter.

johnynek:
I realize we didn't add `MapBoundedSemilattice` when we have `Semilattice` which we could.

Also, there are many similar constructions related to `Option[T]` which is like a `Map[Unit, T]`.

std/src/main/scala/algebra/std/map.scala
TomasMikula: I consider `k -> 0` equivalent, for all purposes, to not having a mapping for key `k` at all. So I don't really care whether zeros are kept or not, except for maybe saving some space.

If we distinguish between `k -> 0` and "no value for key `k`", the behavior is unlawful. (I was actually wondering how is it possible that the tests for e.g. `MapRng` don't fail and then I noticed the `Eq` instance is smart enough to ignore zero values.)

In this setting, the user should only ever call `getOrElse(k, zero)` and then there will be no surprises.

I can imagine a case where it is tempting to encode some additional information in `k -> 0` vs. "no value for `k`", but I'm not sure it is a good idea in combination with `Rng` or `GenBool` operations.

Also note that an explicit call to remove zeros is not possible in a generic setting, where you only have a type class instance operating on some generic values.

std/src/main/scala/algebra/std/map.scala
TomasMikula: I think I tried that, but somehow the type inference wasn't happy. Will check again.

TomasMikula:
I will rewrite the expressions to invoke `iterator` early. Now I'm thinking, if the two maps differ in size by 3 orders of magnitude, it might be better to apply the small number of individual updates on the bigger map.

TomasMikula:
Yeah, I realized there could be interesting structures weaker than `GenBool` for `Map`.

TomasMikula:
I don't know how immutable `Map` is implemented, but if I assume that updating a single key costs log _n_, and if the other (smaller) map has size _m_, then if

_m_ log _n_ < _n_ + _m_

it is more efficient to update the big map.

std/src/main/scala/algebra/std/map.scala
johnynek: I think the absence of the `.` actually changes the order in which scala infers types. Except for operators, I always prefer the `.` for this reason.

johnynek:
Immutable map is a HashTrie, so it is ~ `log_32 n` cost to update a map of size `n`.

johnynek:
@TomasMikula you might want to take a look at: https://github.com/non/algebra/blob/a32f76aba93245b95972b4b4bb80f29b0621c9ba/std/src/main/scala/algebra/std/Util.scala#L154 the addMap/subtractMap methods may be able to be used in these cases.

TomasMikula:
Yeah, `addMap` could be used to implement `or` right away. `subtractMap` could possibly be used to implement `without`, passing constant function returning `zero` as negate (because here, unlike in rings, `0\a = 0`). The problem is that if in `x\y`, `x` is much smaller, `x\y` can be fast, because we don't have to introduce 0s for all the elements of `y` (whereas in `subtractMap` you have to introduce the inverses).

std/src/main/scala/algebra/std/map.scala
TomasMikula: I don't know what I was doing before, but type inference seems to work just fine now.

std/src/main/scala/algebra/std/map.scala
johnynek: I guess this passes because `Eq[Map[K, V]]` is the `MapVectorEq`, and since the guess there is an `AdditiveGroup[Set[Byte]]` somewhere? It's a bit unclear to me.

I wonder if it makes sense that this class also extend or have a `Eq[Map[K, V]]` which makes it clear that `GBV.zero` is the some as a missing value. I get really worried that we have action at a distance here: laws that require that the Eq instance in sync.

TomasMikula: Yes, `MapVectorEq` is the equivalence in use, and there is an instance of `AdditiveMonoid[Set[Byte]]` (not group), namely the `SetSemiring` in `std/set.scala`.

Btw, it looks like `MapEq` is not used anywhere, just `MapVectorEq`.

Regarding whether some type classes should define their own `Eq`, I was wondering the same, but @non [thought otherwise](https://gitter.im/non/cats?at=56423db104a883da3838e0bf), and I tend to agree now.

johnynek: What about it @non? What if this has a member `Eq` or extends `Eq`?

johnynek: I was hoping @non would weigh in here. I don't mean to add to your load, sir.

johnynek:
i think we need the `TotalMap[K, V]` to handle this. @TomasMikula I wonder if your https://github.com/TomasMikula/hasheq could be the right place to add such a type?

TomasMikula:
> i think we need the TotalMap[K, V] to handle this.

I'm afraid you're right.

> I wonder if your https://github.com/TomasMikula/hasheq could be the right place to add such a type?

Indeed, it would make sense to add `TotalMap` to `hasheq`. Then have a third module that provides `algebra` instances for `hasheq` structures.

non:
That sounds like a great plan to me!

TomasMikula:
See
 - TomasMikula/hasheq#8
 - TomasMikula/hasheq#9

PR #128, opened 2015-12-05
Remove algebra.number

tixxit:
There was some discussion in gitter, but the general gist is that

 * algebird and cats both don't use any of the type classes in `algebra.number`
 * spire has since expanded on these type classes to make them more useful, *but*
 * the type classes in spire require number types from Spire and cannot be ported over

I think the best bet for now is to remove the type classes. If we find we want some parts of them later on, we should think of a way of pulling them back out without disrupting the utility of Spire's versions.


rklaehn:
:+1:

Should we do this before or after the initial "spire goes algebra" release?

non:
:+1: I think we should do this before.

johnynek:
Looks good.

Issue #123, opened 2015-11-21
Allow opting out of serializable checks


ceedubs:
Many laws (such as the group laws) force a serializability check. I understand that this can be helpful for making sure your type class instances are Spark/Scalding/etc friendly. However, it would be nice if there were a way to opt-out of this (or even make it opt-in), since it's not always possible to create a serializable instance.

For example, `Future[A]` has a law-abiding `Monoid` instance as long as `A` does, but currently we can't run the monoid laws on it, because the instance captures an `ExecutionContext`, which is not `Serializable`.

johnynek:
I was the one that pushed for this. I can see your point. There are actually two questions:

1. Do we remove the `extends Serializable`
2. Do we remove the test?

Without the marker serializable, Java won't even try to serialize, but even if you mark something as serializable it may fail. Both of these failures generally happen at runtime and almost never at compile time.

I'm 100% okay with an option to opt out in the test.

Even though it is a lie, I would probably prefer to keep the `extends` on the traits because of how easy it is to forget it and how frustrating it is to have to publish a new jar just to add an extends to a dependency.

Serialization fundamentally has runtime failure possibility, so I'm okay with unusual cases causing exceptions.

non:
I pretty much agree with @johnynek. I think instances *should* default to serializable. `Future` is already a bit of a black sheep in some ways, so I think opting-out of the serializability test for it makes the most sense.

ceedubs:
I'm okay with the proposed plan. Since you get runtime as opposed to compile-time failure if you try to serialize an unserializable class whether or not it's marked with `Serializable`, I don't see a lot of point in worrying about whether a `Serializable` trait is inherited.

So I guess then the question becomes how will it look to have tests where you can opt out of serializability? A flag passed to the test constructor/method?

johnynek:
I can think of a few ways:

1. a flag which defaults to checking (hopefully not a boolean, but for instance a simple ADT).
2. a trait to "unmark" serializability. If the item extends that trait, we don't check it. A little weird to extend serializable, and then be unserializable, but this approach allows others to possibly check that you opted out.
3. a typeclass approach for the above. The default is that there is no instance for `NotSerializable[T]` when `T <: Serializable`, but in the companion object your your method you could do this. Problem with this is that we usually use subtyping for typeclasses, and you lose the exact type. I guess this approach is not so great.

ceedubs:
Thank you @johnynek. To me #1 sounds the most appealing.

[...]

PR #129, opened 2015-12-09
add implicit ADT for skipping serialization tests

sritchie:
First stab at #123.


laws/shared/src/main/scala/algebra/laws/RunSerializationLaws.scala
```diff
@@ -0,0 +1,9 @@
+package algebra.laws
+
+sealed trait SerializationCheck
+case object RunSerialization extends SerializationCheck
+case object SkipSerialization extends SerializationCheck
+
+object SerializationCheck {
+  implicit val check: SerializationCheck = SkipSerialization
```
johnynek: I don't think we want to not check by default.

laws/shared/src/main/scala/algebra/laws/RunSerializationLaws.scala
```diff
@@ -0,0 +1,9 @@
+package algebra.laws
+
+sealed trait SerializationCheck
```
johnynek: what if this is `SerializationCheck[M]` so we can default to doing `Run...` but override with `Skip...`.

laws/shared/src/main/scala/algebra/laws/RunSerializationLaws.scala
sritchie: yeah, my bad. Pushed early.

laws/shared/src/main/scala/algebra/laws/RunSerializationLaws.scala
sritchie: Why do we need the type parameter? I had goofed and set it to `Skip` above during some cycling when running the tests, just to make sure that this still took effect. Switching to `Run` by default works great.

johnynek: With the type parameter, in the companion object of `M` you could put whether to check the serialize check or not. Might be a good idea.

But if we call it `SerializationCheck` I think I'd rather call it `IsSerializable[M]` as a typeclass that is the actual promise that `M` is java serializable and it will always work. Then you could make combinators for containers which are basically the `and` on all of these.

I don't know if we need that level of geekout though. Basically the use case we care about are things like Futures mostly.

sritchie: yeah, that makes sense for java serialization. It's odd to have this annotation as well as the java.io.Serializable extension, since the only override you're going to do is mark a class NOT serializable, with `IsNotSerializable[M]`, while the class is still going to be serializable as far as java's concerned.

Letting the class itself opt out seems like the right way to go. Tests and users that want to store instances can have a better check then the current reflection option.

sritchie:
@johnynek changed the name and added the type parameter. Do we still want to thread implicit params through all of the test methods if we're only supporting an actual definition of `IsSerializable[T]` inside the companion object? The only reason to do that is if we want to allow scoping inside the tests. Might be good to just leave it alone.

laws/jvm/src/main/scala/algebra/laws/Platform.scala
```diff
@@ -7,28 +7,33 @@ import Prop.{False, Proof, Result}
...
-      } finally {
-        oos.close()
-        if (ois != null) ois.close()
-      }
-    }
+  def serializable[M: IsSerializable](m: M): (String, Prop) =
+    "serializable" -> (implicitly[IsSerializable[M]] match {
+      case Is() =>
+        Prop(_ => Result(status = Proof))
+      case IsNot() =>
```
ceedubs: Shouldn't we only be checking the serializability if it `Is` serializable instead of if it `IsNot`? :)

laws/js/src/main/scala/algebra/laws/Platform.scala
```diff
@@ -7,9 +7,10 @@ import Prop.{False, Proof, Result}
 
 private[laws] object Platform {
 
-  // Scala-js does not implement the Serializable interface, so we just retuen true.
+  // Scala-js does not implement the Serializable interface, so we just return true.
   @inline
-  def serializable[M](m: M): (String, Prop) = "serializable" -> Prop { _ =>
-    Result(status = Proof)
-  }
+  def serializable[M: IsSerializable](m: M): (String, Prop) =
```
ceedubs: This seems a little bit confusing to me, because it looks like `M` must be serializable in order to call this, when in fact the `IsSerializable` instance could be `IsNot`. I'm at a loss for names that seem any better though.

laws/jvm/src/main/scala/algebra/laws/Platform.scala
sritchie: Ugh, sorry about that. Second time I goofed those cases. Fixed.

ceedubs:
I'm not sure I quite understand https://github.com/non/algebra/pull/129#issuecomment-163461243. I wouldn't think that this approach would require the instance to be defined in the companion object. Would it? In most cases I don't think we'll have the instance in the companion object because 1) Future is the instance where we've wanted this and 2) tests are probably more likely to rely on the laws module than the main code where classes are defined.

johnynek:
I think we have to thread the implicit all the way through to take this approach (even if there is an `IsSerializable` instance for the type) due to type erasure.

I haven't really looked at the hierarchy to see if that will be a disaster.

I agree now with @ceedubs even though I suggested the name. The problem is this is purely type-level function. We aren't going to do anything with it except skip a test at runtime. One way out of the naming thing would be to use `Priority`. We could do `implicit p: Priority[DummyImplicit, NotSerializable[M]]` and then only define `NotSerializable[Future[T]]` for instance. But really, that's just another way to express the ADT in a possibly more indirect way.

I would suggest, to see if this approach works, we set aside the naming question and see how clean it is to plumb it all the way through and have a test for a non-serializable class (maybe something that explicitly throws if it is serialized).

ceedubs:
@johnynek I think that's a good suggestion. And I know understand why you were saying it would need to be threaded all the way through. I had forgotten that other tests call the serializability test, but that's of course why this needed in the first place :)

laws/shared/src/main/scala/algebra/laws/GroupLaws.scala
```diff
@@ -22,103 +22,133 @@ trait GroupLaws[A] extends Laws {
...
+      base = semigroup(A.additive, semiCheck),
+      parents = Nil,
+      Rules.serializable(A),
+      Rules.repeat1("sumN")(A.sumN),
+      Rules.repeat2("sumN", "+")(A.sumN)(A.plus)
+    )
+
+  def additiveCommutativeSemigroup(implicit A: AdditiveCommutativeSemigroup[A],
+    check: IsSerializable[CommutativeGroup[A]],
+    addCheck: IsSerializable[AdditiveCommutativeGroup[A]]) =
```
sritchie: lots of required implicits.

sritchie:
@johnynek, @ceedubs - take a look and see if you guys can stomach the required changes in `LawTests`.

Also - it looks like to reuse these tests, users are going to extend `LawTestsBase`. Currently that means they inherit all of the `algebra` tests, in addition to the helpers. Do we want to move the actual tests out into their own trait, separate from `laws` etc?

laws/shared/src/main/scala/algebra/laws/IsSerializable.scala
```diff
@@ -0,0 +1,17 @@
+package algebra.laws
+
+sealed trait IsSerializable[+T]
```
johnynek: Should this be covariant? I don't think so. For instance, if `Int` has a given `IsSerialializable` instance, I can't in general use that same thing for `Any`.

On the other hand, if I want to test a given type `M1` which I know is a subclass of `M2`, then if I have a `Gen[M1]` that I use for all the tests for `M2` (note Gen is covariant), then in that case the `IsSerializable` instance will work to "upcast". I don't immediately see how to formalize that in the type system. Maybe something like:

```scala
  def upcast[M2, M1 <: M2](g: Gen[M1], isser: IsSerializable[M1]): (Gen[M2], IsSerializable[M2]) =
  (g, isser.asInstanceOf[IsSerializable[M2]])
```
But that is only safe if you use the IsSerializable instance with the Gen. It doesn't really glue them together.

laws/shared/src/main/scala/algebra/laws/RingLaws.scala
```diff
@@ -37,103 +37,144 @@ trait RingLaws[A] extends GroupLaws[A] {
...
+      parent = Some(multiplicativeSemigroup),
+      Rules.repeat0("pow", "one", A.one)(A.pow),
+      Rules.collect0("product", "one", A.one)(A.product)
+    )
+
+  def multiplicativeCommutativeMonoid(implicit A: MultiplicativeCommutativeMonoid[A],
+    check: IsSerializable[CommutativeMonoid[A]],
+    multCheck: IsSerializable[MultiplicativeCommutativeMonoid[A]]) =
+    new MultiplicativeProperties(
+      base = _.commutativeMonoid(A.multiplicative, check),
```
johnynek: cases like this are weird too, since really what you want to do is derive from check. Like if `check == IsNot` then `multCheck` should be IsNot. Maybe just a method to change the type would be fine, and never really unsafe since this is just controlling test behavior.

laws/shared/src/test/scala/algebra/laws/LawTests.scala
```diff
@@ -44,13 +44,22 @@ trait LawTestsBase extends FunSuite with Discipline {
   laws[LogicLaws, Boolean].check(_.bool)
   laws[LogicLaws, SimpleHeyting].check(_.heyting)
   laws[LatticePartialOrderLaws, Boolean].check(_.boundedLatticePartialOrder)
-  laws[RingLaws, Boolean].check(_.boolRing(booleanRing))
+  laws[RingLaws, Boolean].check { l =>
+    implicit val br: BoolRing[Boolean] = booleanRing
```
johnynek: these all change due to the extra implicit parameter?

not awesome. :(

johnynek:
I really appreciate you running this idea down, @sritchie . @ceedubs What do you think? do you have any other ideas? I guess a default parameter that is threaded through is the other idea. Is there any fancy scalacheck trick that we could change an argument within a scope at the scalacheck level so we don't have to litter the code with threading all the way through (that may be ugly in its own way).

sritchie:
Annoying that we can't just add a method at the bottom of the trait heirarchy that supplies the default implicit. If we do that, will any trait that extends the base and overrides that method also pass its new default implicit on to calls to `serializable`?

Then users can just make a new class with an overridden def for types that can't support serialization. Happy to try it. (My mental ability to anticipate implicit scoping's still rusty :)

sritchie:
Ah, I guess not, since `LawTestsBase` doesn't extend that stack.

laws/shared/src/main/scala/algebra/laws/IsSerializable.scala
ceedubs: Agreed that this shouldn't be covariant.

laws/shared/src/test/scala/algebra/laws/LawTests.scala
ceedubs: Yeah, this is the biggest bummer to me, because the other changes aren't particularly pretty, but they are at least hidden within the laws. This affects the user code that wants to check laws.

I'm curious if it would be any cleaner to use the "reader" approach and represent laws as `checkSerializability => A`. The `check` method could be a helper method for when you want serializability to be checked, and there could be an alternative `checkWithoutSerializability` or something that fixes the input to not check serializability. This might help with the changes in this file, since there would no longer be two implicit parameters. It would also help with the places like https://github.com/non/algebra/pull/129/files#r47260083. I'm unsure how clean it would look in general though.

sritchie:
Let's close this one up for now. The changes required for this approach are insane. Let me know what you think of the simpler approach here:

https://github.com/non/algebra/pull/141

Issue #130, opened 2015-12-09
Array instances


rklaehn:
Add instances for arrays.

Arrays are essential in many cases for compact representation, so I think having array instances would be good.

I care most about really basic things like Eq[A] and Order[A], but I guess it would do no harm to just copy everything applicable to algebra from https://github.com/non/spire/blob/master/core/shared/src/main/scala/spire/std/array.scala

denisrosset:
Agree totally.

rklaehn:
Should this use `@specialized` or not?

denisrosset:
Thanks @rklaehn, I guess the issue can be closed!

rklaehn:
Well, it depends. There are a lot of instances that we could provide. I just copied over the essential ones. But I think we should at least provide Monoid before we close the issue.

johnynek:
We probably want a few:

```scala
def concatMonoid[T: ClassTag]: Monoid[Array[T]]
// then the usual vector space style ones, which are probably what people want:
def extendingPointwiseSemigroup[T: ClassTag: Semigroup]: Semigroup[Array[T]]
// you can do pretty much everything this way in core except Field, I think.
// It is a shame we have to write those all out. Maybe macros could help here.
```

PR #132, opened 2015-12-09
Add array instances

rklaehn:
Copied from spire, adjusted to algebra conventions, and reduced to just
Eq/Order for now. Also added law checks for Array[Int]

Implements #130

### DIFF
--- a/laws/shared/src/test/scala/algebra/laws/LawTests.scala
+++ b/laws/shared/src/test/scala/algebra/laws/LawTests.scala
@@ -192,5 +192,6 @@ trait LawTestsBase extends FunSuite with Discipline {
 
   laws[OrderLaws, Int]("fromOrdering").check(_.order(Order.fromOrdering[Int]))
   laws[OrderLaws, Array[Int]].check(_.order)
+  laws[OrderLaws, Array[Int]].check(_.partialOrder)
 }
 
--- a/laws/shared/src/test/scala/algebra/laws/LawTests.scala
+++ b/laws/shared/src/test/scala/algebra/laws/LawTests.scala
@@ -191,5 +191,6 @@ trait LawTestsBase extends FunSuite with Discipline {
   }
 
   laws[OrderLaws, Int]("fromOrdering").check(_.order(Order.fromOrdering[Int]))
+  laws[OrderLaws, Array[Int]].check(_.order)
 }
 
--- a/std/src/main/scala/algebra/std/array.scala
+++ b/std/src/main/scala/algebra/std/array.scala
@@ -1,0 +1,36 @@
+package algebra
+package std
+
+import scala.{specialized ⇒ sp}
+
+package object array extends ArrayInstances
+
+trait ArrayInstances {
+  implicit def arrayOrder[@sp A: Order]: Order[Array[A]] = new ArrayOrder[A]
+}
+
+object ArraySupport {
+
+  def eqv[@sp A: Eq](x: Array[A], y: Array[A]): Boolean = {
+    var i = 0
+    if (x.length != y.length) return false
+    while (i < x.length && i < y.length && Eq.eqv(x(i), y(i))) i += 1
+    i == x.length
+  }
+
+  def compare[@sp A: Order](x: Array[A], y: Array[A]): Int = {
+    var i = 0
+    while (i < x.length && i < y.length) {
+      val cmp = Order.compare(x(i), y(i))
+      if (cmp != 0) return cmp
+      i += 1
+    }
+    x.length - y.length
+  }
+}
+
+private final class ArrayOrder[@sp A: Order]
+  extends Order[Array[A]] with Serializable {
+  override def eqv(x: Array[A], y: Array[A]): Boolean = ArraySupport.eqv(x, y)
+  def compare(x: Array[A], y: Array[A]): Int = ArraySupport.compare(x, y)
+}
--- a/std/src/main/scala/algebra/std/array.scala
+++ b/std/src/main/scala/algebra/std/array.scala
@@ -13,6 +13,11 @@ trait ArrayInstances {
 
 private object ArraySupport {
 
+  private def signum(x: Int): Int =
+    if(x < 0) -1
+    else if(x > 0) 1
+    else 0
+
   def eqv[@sp A: Eq](x: Array[A], y: Array[A]): Boolean = {
     var i = 0
     if (x.length != y.length) return false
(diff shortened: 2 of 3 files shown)


std/src/main/scala/algebra/std/array.scala
```diff
@@ -0,0 +1,36 @@
...
+package object array extends ArrayInstances
+
+trait ArrayInstances {
+  implicit def arrayOrder[@sp A: Order]: Order[Array[A]] = new ArrayOrder[A]
+}
+
+private object ArraySupport {
+
+  def eqv[@sp A: Eq](x: Array[A], y: Array[A]): Boolean = {
+    var i = 0
```
johnynek: is there any benefit to store `x.length` in a local val, or with the jit totally take care of that?

std/src/main/scala/algebra/std/array.scala
```diff
@@ -0,0 +1,36 @@
...
+
+  def compare[@sp A: Order](x: Array[A], y: Array[A]): Int = {
+    var i = 0
+    while (i < x.length && i < y.length) {
+      val cmp = Order.compare(x(i), y(i))
+      if (cmp != 0) return cmp
+      i += 1
+    }
+    x.length - y.length
+  }
```
johnynek: what about PartialOrder?

std/src/main/scala/algebra/std/array.scala
```diff
@@ -0,0 +1,36 @@
+package algebra
+package std
+
+import scala.{specialized ⇒ sp}
+
+package object array extends ArrayInstances
+
+trait ArrayInstances {
+  implicit def arrayOrder[@sp A: Order]: Order[Array[A]] = new ArrayOrder[A]
```
johnynek: I think you want arrayEq and arrayPartialOrder for cases where you don't have a total Order.

johnynek:
this needs to test that we are following the laws here. Also, I'd like to test to be sure it is behaving as we expect (dictionary ordering) (which I think could be tested by comparing against Ordering.Iterable).

std/src/main/scala/algebra/std/array.scala
rklaehn: I am pretty sure a loop like this will be optimised perfectly by the JIT

laws/shared/src/test/scala/algebra/laws/LawTests.scala
```diff
@@ -191,5 +191,6 @@ trait LawTestsBase extends FunSuite with Discipline {
   }
 
   laws[OrderLaws, Int]("fromOrdering").check(_.order(Order.fromOrdering[Int]))
+  laws[OrderLaws, Array[Int]].check(_.order)
```
johnynek: sorry, I overlooked these tests.

rklaehn: They just check that order is consistent, not that it is dictionary ordering. I wasn't sure where to put normal non-law-tests...

std/src/main/scala/algebra/std/array.scala
```diff
@@ -0,0 +1,42 @@
...
+    x.length - y.length
+  }
+}
+
+private final class ArrayEq[@sp(Int,Float,Long,Double) A: Eq]
+  extends Eq[Array[A]] with Serializable {
+  def eqv(x: Array[A], y: Array[A]): Boolean = ArraySupport.eqv(x, y)
+}
+
+private final class ArrayOrder[@sp A: Order]
```
johnynek: why the difference here in `@sp`? Make consistent?

laws/shared/src/test/scala/algebra/laws/LawTests.scala
johnynek: Me neither actually. I guess just make a test and use scala check like normal. Something like:

```scala
forAll { (a: Array[Int], b: Array[Int]) =>
  Order[Array[Int]].compare(a, b) == Ordering.Iterable[Int](a.toList, b.toList)
}
```
should do it.

std/src/main/scala/algebra/std/array.scala
```diff
@@ -0,0 +1,42 @@
+package algebra
+package std
+
+import scala.{specialized ⇒ sp}
+
+package object array extends ArrayInstances
+
+trait ArrayInstances {
+  implicit def arrayEq[@sp A: Eq]: Eq[Array[A]] = new ArrayEq[A]
```
johnynek: these should probably have the same `@sp(Int, Float, Long, Double)` that we standardized on.

rklaehn: I think arrayEq should be specialized on everything. Equality comparison of e.g. byte arrays should not have boxing overhead.

For PartialOrder, just `@sp(Int, Float, Long, Double)` seems OK.

johnynek: The motivation was that byte boxing, which uses an array lookup table, was much faster than int boxing (which allocates), so it wasn't worth the extra methods.

But I guess `Eq` is still `@sp`, to keeping thes one `@sp` is fine. Actually, I guess `PartialOrder` and `Order` are `@sp` too, so, that's fine to stay consistent with them.

I think for Semigroup and it's subclasses we went with just the 4 and 8 byte types.

rklaehn: I don't really like the fact that an array comparison can return values other than 0, -1, +1. Want me to change that? Also, want me to squash the commits once we have something that we're happy with?

laws/shared/src/test/scala/algebra/laws/LawTests.scala
rklaehn: Where?
`std/src/test/scala/algebra/std` or `laws/src/shared/test/scala/algebra/std`

johnynek: I guess `std/src/test/scala/algebra/std`

std/src/main/scala/algebra/std/array.scala
johnynek: I'm fine with such a change. I don't care (or particularly like) squashed commits. I prefer to keep the whole history.

std/src/main/scala/algebra/std/array.scala
```diff
@@ -0,0 +1,60 @@
...
+    }
+    x.length - y.length
+  }
+
+  def partialCompare[@sp A: PartialOrder](x: Array[A], y: Array[A]): Double = {
+    var i = 0
+    while (i < x.length && i < y.length) {
+      val cmp = PartialOrder.partialCompare(x(i), y(i))
+      // Double.NaN is also != 0
+      if (cmp != 0) return cmp
```
johnynek: can we write `cmp != 0.0`?

johnynek:
:+1:

PR #134, opened 2015-12-09
Use diff instead of -- for set difference

rklaehn:
At least in case of scala.collection.immutable.HashSet, diff is a tree/tree
operation, whereas -- removes all elements of rhs from lhs one by one. So
diff is much more efficient.

I did this change (tree/tree operations) to s.c.i.HashSet for 2.11. I also think that it would be a good idea to have xor defined in s.c.i.Set for 2.12.

https://github.com/scala/scala/blob/v2.11.7/src/library/scala/collection/immutable/HashSet.scala#L1


TomasMikula:
I wrote the code with `--`. I didn't know it's not the same as `diff`. Thanks!

:+1:

rklaehn:
Really, -- should just call diff in HashSet. But we did not dare to do it because of backwards compatibility in some really weird cases :-(

Issue #135, opened 2015-12-09
Move algebra.std.Rat to tests


rklaehn:
It seems that algebra.std.Rat is there mostly for testing some typeclasses. 

I think if people want a rational number type for production use, they would be much better off with spire.math.Rational. In any case, algebra should not contain data structures.

So I think Rat belongs into the test scope. Otherwise people *will* use it in production.

tixxit:
:+1:

PR #136, opened 2015-12-09
Moved Rat into algebra.laws test scope

rklaehn:
Also moved Rat arbitrary directly into Rat companion object. CheckSupport
is now empty. Remove?

Implements #135


johnynek:
:+1:

tixxit:
:+1:

PR #139, opened 2015-12-18
Upgrade to sbt 0.13.9 & remove some boilerplate

dwijnand:


johnynek:
I'm not an SBT guru, so I'd appreciate a review from someone else here.

johnynek:
strictly speaking, we have more code after doing this boilerplate removal, so I'm not immediately sold that the abstraction is worth it, since people have to lookup where these `.ifScala210` enrichments are coming from.

dwijnand:
That's true, but they're both unfortunate implementation details of not being able to define extension methods in sbt files (to which I've received no pushback on from @jsuereth when I spoke to him about it, it's just not currently available) and the verbose boilerplate required for defining extension methods in the first place.

Initially I wanted to add the ticket number for why `-Ywarn-value-discard` is disabled, then I realised it only need be disabled for 2.10. But I didn't want to _add_ it for only 2.11, I wanted to disable it for 2.10, same as `-Ywarn-unused-import`. Then I wanted to remove the noisy CrossVersion.partialVersion pattern matching, present in two different places...

I felt like `.ifScala210` conveyed the intent, but I'm happy to change that for better alternatives. I'm also very "pro-extension methods", but I'm happy to change that here if a function is preferred.

Issue #131, opened 2015-12-09
Names of operations of for semigroups, monoids, groups


denisrosset:
In algebra, the name of the associative operation is `combine`, the identity is `empty`, the identity test is `isEmpty`, while in spire, it is `op`, `id` and `isId`.

The `combine/empty` terminology is appropriate when concatenating data structures, while `op/id` is standard abstract algebra.

The change came from non/algebra@e0ccca1 (thanks @rklaehn for finding it), and the compromise came probably from usage in cats --- but I was unable to find the discussion leading to the commit.

When basing spire on the algebra typeclasses, this notation is going to impact especially the people using `Group` in an algebraic setting.

What should be done?

I see two options:
1. standardizing on one set of terminology, and have syntax adapters in client libraries to provide the missing names (e.g. in spire or cats, depending on the context)
2. having both terminologies in the typeclass

johnynek:
In fact algebra considerably predates cats. Algebra started as a project
for a common core to spire and algebird.

In algebird we made the choice of Ring extending Group, and made the choice
to just call the semigroup operation plus (yes it is sometimes weird, but
the most common instances make sense and being less abstract there I
believe helped adoption within Twitter).

For algebird I had planned (but not thought too deeply) about making
algebird Semigroup have final combine that calls plus. Thus if you subclass
an algebird Semigroup your code does not change but it works as an algebra
Semigroup.

Having adapters is another approach but it makes it less convenient when it
is time to interop, but he upside there is that you could put them in a
separate package and those that don't want them don't have to have the
dependency (I'm paranoid about dependencies from the trauma of working in a
monorepo).

denisrosset:
Thanks, I did not see the `empty` / `combine` methods in algebird.

johnynek:
They have not been added. I think empty/combine actually came from scalaz
back we we had planned to use these classes there as well.

But my current (not well thought out) plan is to add the methods.

On Wednesday, December 9, 2015, Denis Rosset <notifications@github.com>
wrote:

> Thanks, I did not see the empty / combine methods in algebird.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/non/algebra/issues/131#issuecomment-163318705>.
>


-- 
P. Oscar Boykin, Ph.D. | http://twitter.com/posco | http://pobox.com/~boykin

TomasMikula:
> I think empty/combine actually came from scalaz

[...]

PR #140, opened 2015-12-20
Renamed 'empty' to 'neutral' and 'isEmpty' to 'isNeutral'.

denisrosset:
According to discussion in non/algebra#131

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ prefix representation of the unary function `g`, and `a` is a value
 
 The most basic structures can be found in the `algebra` package. They
 all implement a method called `combine`, which is associative. The
-identity element (if present) will be called `empty`, and the inverse
+identity element (if present) will be called `neutral`, and the inverse
 method (if present) will be called `inverse`.
 
 |Name                |Associative?|Commutative?|Identity?|Inverse?|Idempotent?|
--- a/core/src/main/scala/algebra/Eq.scala
+++ b/core/src/main/scala/algebra/Eq.scala
@@ -109,7 +109,7 @@ object Eq extends EqFunctions {
    * checks that all equality checks pass
    */
   def allEqualBoundedSemilattice[A]: BoundedSemilattice[Eq[A]] = new BoundedSemilattice[Eq[A]] {
-    def empty = allEqual[A]
+    def neutral = allEqual[A]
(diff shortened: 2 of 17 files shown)


johnynek:
I want to make sure @non is +1.

I think @TomasMikula is (he suggested the change) as is @rklaehn, which along with @non seems like a quorum of the most active contributors/users.

TomasMikula:
+1

denisrosset:
@non, can we merge this ?

kailuowang:
This would be a major breaking change for cats - I suspect that a large portion of cats users( me for sure) will have to change their code. How about we not include this in the next release which will block 0.5.0 cats release?

denisrosset:
@kailuowang agree that the change is going to break existing code. That will be the case for me on the spire side if spire is merged back with algebra (around 20kloc to be precise).

However, the longer we wait, the more painful it's going to be for users of both libraries.

rklaehn:
Agree with @denisrosset here. If we want interoperable typeclass based libraries, we need to agree on names. This was discussed at length, and everybody seemed to be OK with the choices. So now algebra just needs to do it.

The alternative is everybody doing his own thing, and the scala ecosystem not being viable for typeclass-based libraries. There is way too much fracturing already.

johnynek:
Two ideas:

1) add an implicit enrichment inside `object Monoid` so that all `Monoid` instances get `.empty` that calls `.neutral`.

2) just suck it up and keep empty. No library is perfect and there kinds of changes push costs on users. They slow adoption of new versions and justify the fracturing that @rklaehn is speaking of.

I want better shared libraries for scala (I think it may mean we need to build source dependency mechanisms into sbt and those becoming the default, but another story).

I'd vote for 1, I'd support 2. I don't want to jam up users.

denisrosset:
I'd vote for 2)

Let's have a default consistent terminology (`empty`, `combine`, `remove`), instead of the half-baked `neutral`.

And let's add enrichment methods in Spire.

non:
@denisrosset Sounds good to me. Thanks all!

non:
To be clear: we will keep `empty` and `combine`, and standardize Spire to those names (with aliases to our current names or vice versa).

non:
(Also, I'm sorry that I have been absent from these conversations over the last few months.)

PR #143, opened 2016-03-20
Eq[String]

adelbertc:


johnynek:
isn't `Order` a supertype? Why is this needed?

adelbertc:
Why yes.. yes it is :|

Issue #148, opened 2016-04-21
"Vector" instances are questionable


johnynek:
@non @tixxit and I were talking about `mapVectorEq` and for instance the `Group` on `Map[K, V]` that assumes that a missing key is the same as `Group.empty`.

This has some issues. Without the correct `Eq[Map[K, V]]` in scope, you violate the laws. Moreover, the standard `Eq` should not be one that has this behavior, because that does not make sense for a lot of use cases.

Our solution is to wrap such instances:
```scala
// infinite map defined at all keys K
trait SparseMap[K, V] {
  def get(k: K): V
  def updated(k: K, v: V): SparseMap[K, V]
  def toMap: Map[K, V]
}

// an infinite vector that is defined at all points (i: Int)
trait SparseVector[T] {
  def apply(i: Int): T
  def updated(i: Int, v: T): SparseVector[T]
  def toVector: Vector[T]
}
```
In this way #126 would be an instance on `SparseMap`

TomasMikula:
The solution by defining a new type for these versions of `Map` and `Vector` sounds good. What is the motivation behind naming them `Vector*`?

adelbertc:
@TomasMikula I believe the `Vector` prefix is in reference to vector equality like in linear algebra

TomasMikula:
Oh, I see, but you mean _infinite-dimensional_ vectors, right? Then a more descriptive name would be `Infinite{Vector|Map}` instead of `Vector{Vector|Map}`, wouldn't it?

tixxit:
I think @johnynek mentioned SparseMap as a possibility, which is descriptive too. In the case of `Vector` it's "really" a vector with Int.MaxValue elements too :\

TomasMikula:
"Sparse" sounds good!

adelbertc:
👍 Sparse

rklaehn:
On the issue of names: I got a library https://github.com/rklaehn/abc where I have implementations of Seq and Map that allow defining a lot of typeclass instances without having a strange `Eq`. I called them `TotalArrayMap` / `TotalArraySeq`. Whether they are sparse or not is an implementation detail IMHO. The important thing is that in both cases the apply method is total, because there is a default value in case no mapping is provided.
https://github.com/rklaehn/abc#-totalarraymapk-v

But sparse is also fine. Just pease don't have a strange `Eq` on the default implicit scope just to make normal map work...

TomasMikula:
[...]

PR #152, opened 2016-04-30
Add cats-kernel dependency.

non:
This commit switches the algebra-core package over to just
aliasing the cats-kernel type classes and companion objects.
It also uses cats.kernel.std instances instead of the
corresponding algebra.std instances.

It also updates the scala.js version to be compatible with
cats.kernel, and removes some unnecessary imports.


non:
This is currently using a local snapshot. I plan to publish cats 0.6.0-M1 and update this PR to that dependency before asking for a merge. But I wanted to put this out there so folks could take a look at it.

laws/shared/src/test/scala/algebra/laws/LawTests.scala
```diff
@@ -107,9 +108,9 @@ trait LawTestsBase extends FunSuite with Discipline {
   laws[RingLaws, Set[String]].check(_.semiring)
 
   laws[OrderLaws, Map[Char, Int]].check(_.eqv)
-  laws[RingLaws, Map[Char, Int]].check(_.rng)
```
johnynek: did these break because of the change to the `Eq`?

project/plugins.sbt
```diff
@@ -4,5 +4,5 @@ addSbtPlugin("com.jsuereth"      % "sbt-pgp"         % "1.0.0")
 addSbtPlugin("org.xerial.sbt"    % "sbt-sonatype"    % "0.5.0")
 addSbtPlugin("com.typesafe.sbt"  % "sbt-ghpages"     % "0.5.3")
 addSbtPlugin("com.typesafe.sbt"  % "sbt-site"        % "0.8.1")
-addSbtPlugin("org.scala-js"      % "sbt-scalajs"     % "0.6.5")
+addSbtPlugin("org.scala-js"      % "sbt-scalajs"     % "0.6.8")
```
johnynek: I think `0.6.9` is out now, right?

johnynek:
looks good to me.

One comment: I guess we should never add an instance here to any type not defined here if it is only for typeclasses in cats (e.g. Semigroup, Monoid, Group, etc...)

So, for the #148 discussion, we might consider adding the `TotalMap`, `TotalVector` type to cats kernel? Or just since we added it here, it is safe to add the instances too, of course.

rklaehn:
@johnynek

I would be against having *any* sort of data structure in cats-kernel, even if it is just a trivial wrapper around `scala.collection.immutable` data structures. It should be very small and just contain completely uncontroversial typeclasses.

Regarding the implementation, I think TotalX shouldn't have much except an apply method and an O(1) way to get the underlying non-total map/vector.

laws/shared/src/test/scala/algebra/laws/LawTests.scala
non: I removed the RNG instance for `Map` because of the changes to `Eq[Map[K, V]]`.

project/plugins.sbt
non: Yeah looks like it just came out. I'll bump the version soon.

non:
As far as "vector" wrappers go -- yeah, I think that is a good starting principle.

I think having *algebra* define simple "vector" wrappers makes sense. I think @rklaehn is right that we should have (unsafe) accessors, but we should also provide a minimal interface to use the type as a vector. The reason I think accessing the underlying structure is (potentially) unsafe is that we want to treat things like `Map(1 -> 0.0)` and `Map.empty[Double]` as equivalent, but the laws for `Eq` mandates that they behave identically.

rklaehn:
@non, the way I solve it in https://github.com/rklaehn/abc is that in `TotalArrayMap`, I never store mappings to the default value. It's my fetish with canonical representations. Without that property, you wouldn't be able to use structural comparison to compare TotalArrayMap instances, which would suck.
```
scala> ArrayMap(1 -> 1).withDefault(1)
res1: com.rklaehn.abc.TotalArrayMap[Int,Int] = TotalArrayMap((), 1)
```
A consequence is that sometimes, `x.withDefault(y).withoutDefault =!= x`, which might be unexpected.

non:
@rklaehn Right, that is definitely the correct way to do it.

non:
Assuming Travis is happy, I think we should merge this. It depends on a stable Cats milestone (0.6.0-M2) and will allow us to make this transition now anticipating Cats 0.6.0.

core/src/main/scala/algebra/package.scala
```diff
@@ -1,2 +1,38 @@
 package object algebra {
+
```
johnynek: would it be too much of a pain to sort the types alphabetically.

std/src/main/scala/algebra/std/set.scala
```diff
@@ -6,9 +6,8 @@ import algebra.ring.{BoolRng, Semiring}
 
 package object set extends SetInstances
 
-trait SetInstances {
+trait SetInstances extends cats.kernel.std.SetInstances {
   implicit def setLattice[A]: GenBool[Set[A]] = new SetLattice[A]
-  implicit def setPartialOrder[A]: PartialOrder[Set[A]] = new SetPartialOrder[A]
   implicit def setSemiring[A] = new SetSemiring[A]
   def setBoolRng[A] = new SetBoolRng[A]
```
johnynek: why is this not implicit? Can we comment?

project/plugins.sbt
johnynek: should we bump this now? why bump from 0.6.5 to 0.6.8 but not 0.6.9?

johnynek:
some minor comments, but pretty much ready to merge.

std/src/main/scala/algebra/std/set.scala
TomasMikula: For the record, because it is not compatible with the above `Semiring` (they differ in the definition of `plus`, "union" vs. "symmetric difference"). And I see that is not obvious at all. We should document the meaning (in terms of `Set` operations) of Ring operations.

TomasMikula: And I'm happy to add these comments after this PR is merged, so as not to ask @non to fix deficiencies introduced by me.

non:
I'll go ahead and handle the obvious issues @johnynek pointed out.

TomasMikula:
👍

johnynek:
👍

milessabin:
:tada:

PR #155, opened 2016-07-28
typo: Rig -> Ring

dckc:


non:
This isn't a typo. If you look you'll see `CommutativeRing` below. We have support for both *rigs* and *rings*.

johnynek:
so good! I love this PR. I hope we keep getting it!

(also, `Rng` fixes welcome).

dckc:
Oh. Learn something new every day. Wikipedia says "The term rig is also used occasionally ..." under Semiring. Any particular reason not to use "semiring"?

dckc:
Ah. semiring has 1.

tixxit:
Yeah - we use Semiring to mean a Ring without an inverse or a multiplicative identity. I think that Semiring is often used for both these cases, with the context in which it's used disambiguating it. However, we need both so we disambiguate them all with Semiring (no multiplicative identity, no inverse), Rig (no inverse), Rng (no multiplicative identity), and Ring :)

PR #156, opened 2016-08-18
Bump Cats to 0.6.1 (and Scala.js to 0.6.11).

non:
This also deletes an erroneous semilattice that we were using for law
tests (but not shipping to anyone else). It turns out not to be
associative.


johnynek:
👍

PR #157, opened 2016-08-21
Improve the performance of Map's algebras.

non:
We noticed an issue in cats-kernel where operations over Maps were
incredibly slow. This commit ports those fixes over here, where our
additive/multiplicative monoids experienced the same issues.

Review by @johnynek

### DIFF
--- a/std/src/main/scala/algebra/std/map.scala
+++ b/std/src/main/scala/algebra/std/map.scala
@@ -1,8 +1,10 @@
 package algebra
 package std
 
-import cats.kernel.std.util.StaticMethods.addMap
-import algebra.std.util.StaticMethods.timesMap
+import scala.collection.mutable
+
+import cats.kernel.std.util.StaticMethods
+
 import algebra.ring._
 
 package object map extends MapInstances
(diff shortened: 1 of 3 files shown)


std/src/main/scala/algebra/std/map.scala
```diff
@@ -30,23 +32,98 @@ class MapAdditiveMonoid[K, V](implicit V: AdditiveSemigroup[V]) extends Additive
...
+        })
+      }
+    }
+
+  override def sumN(a: Map[K, V], n: Int): Map[K, V] =
+    if (n > 0) a.map { case (k, v) => (k, V.sumN(v, n)) }
+    else if (n == 0) zero
+    else throw new IllegalArgumentException("Illegal negative exponent to sumN: %s" format n)
+
+  override def sum(as: TraversableOnce[Map[K, V]]): Map[K, V] = {
```
johnynek: we don't have `trySum` here:

https://github.com/non/algebra/blob/master/ring/src/main/scala/algebra/ring/Additive.scala#L32

can we do the same override on AdditiveMonoid or here so that `trySum` calls `sum`?

johnynek:
minor comment about `trySum`.

std/src/main/scala/algebra/std/map.scala
non: Sure.

std/src/main/scala/algebra/std/map.scala
```diff
@@ -30,23 +32,98 @@ class MapAdditiveMonoid[K, V](implicit V: AdditiveSemigroup[V]) extends Additive
...
 
 class MapSemiring[K, V](implicit V: Semiring[V]) extends MapAdditiveMonoid[K, V] with Semiring[Map[K, V]] {
 
   def times(xs: Map[K, V], ys: Map[K, V]): Map[K, V] =
+    // we figure out which of our maps is smaller, iterate over its
+    // keys, see which of those are in the larger map, and add the
+    // resulting product to our result map.
     if (xs.size <= ys.size) {
-      timesMap(xs, ys)((x, y) => V.times(x, y))
+      xs.foldLeft(Map.empty[K, V]) { case (m, (k, x)) =>
```
johnynek: I wonder if `Map.empty` makes sense here or if we should use the mutable approach? Like, we could allocate an empty `MMap` about the size of the smallest entry (an upperbound).

I *think* this will be faster, if we care, but we can come back to it later if you prefer.

non: Let's do this now, and then make that change with a benchmark later.

johnynek:
👍

PR #158, opened 2016-08-21
Topic/kernel 070 update

non:
This PR is dependent on #157.


johnynek:
this looks fine to me (modulo) the fix in for Map algebra.

Issue #160, opened 2016-08-22
Proposal: merge lattice, ring and instances


johnynek:
Since core was basically converted to be `cats-kernel` there is basically nothing there.

I think there are several reasons for modules:

1. jar size minimization for a la carte usage
2. dependency minimization when some dependencies are only used by a subset of features
3. different stability guarantees (some jar may be experimental).

I don't really see any of those applying anymore with kernel in cats.

I think we should strive for mima-checked stability (turned on CI), but we should merge the projects.

What to call the jar is tricky. Moving classes between artifacts can give some tricky maven issues. In this case, we may move between the same outer project, so maybe it is okay to put them all in algebra-core, since the previous jars depended on core anyway (this is a kind of migration within the existing dependency graph). I think it would be okay to merge all into core.

Any objections to this?

non:
That makes sense to me.

PR #161, opened 2016-08-29
Remove various unnecessary modules

striation:
This PR implements the proposal in #160. It changes the module structure without changing any of the core type classes (although it does remove two types used in testing: `TypeTagM` and `Platform`, which are now available to us through our *catalysts* test dependency).

After this change, we would have only two subprojects:

 - *algebra-core*: contains all type classes, etc.
 - *algebra-laws*: contains all laws, law-checking code, tests and helpers, etc.

I didn't go ahead and enable MiMA on this yet because I wanted to float the idea of removing those two types that are extraneous. I can readd them if we think compatibility there is worth it. I'm also happy to verify that those two types are the only things that changed.

What do you think @johnynek? Also interested to hear from @tixxit and others on this.


johnynek:
I'm +1 but let's get one more buy in.

tixxit:
With the move to cats kernel, it takes the pressure off to provide such a small surface area. I'm definitely 👍  on this.

PR #162, opened 2016-08-31
Add 0.5.0 release to readme and do some cleanup.

non:
I just pushed 0.5.0 to Sonatype, so this PR will update the version in our README.

I also *linkified* the contributor names, rephrased some paragraphs, and added some code blocks to other parts of the README.

I think 0.5.0 should be a candidate for where we start our MiMA compatibility requirements. Here's my proposal for what happens next:

 1. @non, @tixxit, @rklaehn, and others will get a Spire branch working that uses *algebra-0.5.0*.
 1. We test that Spire branch to be sure that we don't see major problems with our benchmarks.
 1. @johnynek and others optionally do the same thing for Algebird.
 1. If needed, we make changes to *cats-kernel* and *algebra*, re-release, goto (1).
 1. We all agree that we are happy with *cats-kernel* and *algebra* in their current form.
 1. We freeze *cats-kernel* in MiMA compatibility mode.
 1. We freeze *algebra* in MiMA compatibility mode (and call it 1.0.0).
 1. Profit!

(In the meantime we can merge PRs but should aim to keep MiMA compatibility with 0.5.0 unless we notice a serious design flaw that we have to fix.)

What do you all think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,19 +13,28 @@ these type classes, we will distribute an `algebra` package with no
 dependencies that works with Scala 2.10 and 2.11, and which can be
 shared by all Scala libraries interested in abstract algebra.
 
+Since the creation of Algebra, we have also decided to interoperate
+with the [Cats](http://github.com/typelevel/cats) project. Algebra and
+Cats interoperate using the *cats-kernel* module.
+
 ## getting algebra
 
 Algebra supports Scala 2.10 and 2.11, and is available from Sonatype
-(and Maven Central). To use algebra in your own projects, include this
-snippet in your `build.sbt` file:
+(and Maven Central). Algebra is also available for Scala.js.
+
+To use algebra in your own projects, include this snippet in your
+`build.sbt` file:
 
 ```scala
-resolvers += Resolver.sonatypeRepo("releases")
-
-libraryDependencies += "org.spire-math" %% "algebra" % "0.4.0"
+libraryDependencies += "org.spire-math" %% "algebra" % "0.5.0"
 ```
 
-As of 0.3.1 algebra also supports scala.js!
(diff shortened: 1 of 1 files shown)


tixxit:
This all sounds great to me. And the PR looks good too 👍

non:
One other possible change we might want to do pre-freeze is to make our law definitions consistent with Cats'. I think that style is a bit easier to read/understand (since it came later). That idea is tracked at https://github.com/typelevel/cats/issues/1272.

README.md
```diff
@@ -13,19 +13,28 @@ these type classes, we will distribute an `algebra` package with no
...
+(and Maven Central). Algebra is also available for Scala.js.
 
-```scala
-resolvers += Resolver.sonatypeRepo("releases")
+To use algebra in your own projects, include this snippet in your
+`build.sbt` file:
 
-libraryDependencies += "org.spire-math" %% "algebra" % "0.4.0"
+```scala
+libraryDependencies += "org.spire-math" %% "algebra" % "0.5.0"
```
TomasMikula: This now needs to be `org.typelevel`.

README.md
```diff
@@ -13,19 +13,28 @@ these type classes, we will distribute an `algebra` package with no
...
 ```
 
-As of 0.3.1 algebra also supports scala.js!
+If you want to use Algebra's laws, you can include those as well with
+this snippet:
+
+```scala
+libraryDependencies += "org.spire-math" %% "algebra-laws" % "0.5.0"
+```
```
TomasMikula: Likewise here.

README.md
non: Good catch! I'll put this in a PR asap.

PR #165, opened 2016-09-09
Work-around for SI-8431.

non:
If we use `super` in the implementation of a specialized trait it can
cause problems for implementing classes downstream. I hit this bug [1]
when trying to get Spire using Algebra.

[1] https://issues.scala-lang.org/browse/SI-8431


tixxit:
👍

johnynek:
:+1:

PR #168, opened 2016-09-11
Add combineAll to Additive algebras

johnynek:
We missed these, so converting to cats-kernel types would lose the optimizations for sequences.


non:
LGTM 👍

tixxit:
👍

Issue #167, opened 2016-09-09
add code coverage checks


johnynek:
We should start tracking (and increasing) code coverage.

non:
Agreed! I'll try to set this up ASAP.

PR #170, opened 2016-09-13
Add scoverage and report results to codecov.io

tixxit:
I copied the setup from tixxit/delimited - I think this should just ~work. I've also updated the buttons in the README to point to typelevel/aglebra now.


tixxit:
See #167

johnynek:
👍

tixxit:
Good news: we have coverage reports!
Bad news: 68% coverage :(

PR #171, opened 2016-09-14
Add Rüdiger and Tomas to the list of maintainers.

non:
I nominated both of them, the existing maintainers were unanimous in
agreeing to offer maintainership to them, and they both accepted.

I've also increased number of sign-offs needed (N) from 1 to 2.

With seven maintainers we should be able to meet this requirement,
even if some of us are less active than others (for many significant
PRs we've already done this).

If this ends up being too much of a burden we can reduce N back to 1.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,27 +47,31 @@ tracker or on relevant pull requests.
 
 ## participants
 
-Anyone who wants to participate should feel free to send pull requests
-to this repo. The following people have push access:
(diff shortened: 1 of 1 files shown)


tixxit:
:+1:

johnynek:
:+1: would be so great to have them as maintainers! Thank you both.

PR #173, opened 2016-09-28
Add Ring#fromBigInt

tixxit:
This is similar to `fromInt`, but works with larger numbers. There is a default implementation that is based on `fromInt`, but this method is easily implementable by all our instances that also override `fromInt`.

### DIFF
--- a/laws/src/main/scala/algebra/laws/RingLaws.scala
+++ b/laws/src/main/scala/algebra/laws/RingLaws.scala
@@ -107,7 +107,15 @@ trait RingLaws[A] extends GroupLaws[A] {
     name = "ring",
     al = additiveCommutativeGroup,
     ml = multiplicativeMonoid,
-    parents = Seq(rig, rng)
+    parents = Seq(rig, rng),
+    "fromInt" -> forAll { (n: Int) =>
+      A.fromInt(n) ?== A.sumN(A.one, n)
+    },
+    "fromBigInt" -> forAll { (ns: List[Int]) =>
+      val actual = A.fromBigInt(ns.map(BigInt(_)).foldLeft(BigInt(1))(_ * _))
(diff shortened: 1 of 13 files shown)


laws/src/main/scala/algebra/laws/RingLaws.scala
```diff
@@ -107,7 +107,15 @@ trait RingLaws[A] extends GroupLaws[A] {
     name = "ring",
     al = additiveCommutativeGroup,
     ml = multiplicativeMonoid,
-    parents = Seq(rig, rng)
+    parents = Seq(rig, rng),
+    "fromInt" -> forAll { (n: Int) =>
+      Ring.fromInt[A](n) ?== A.sumN(A.one, n)
+    },
+    "fromBigInt" -> forAll { (ns: List[Int]) =>
```
johnynek: why is this not
```scala
forAll { bi: BigInt =>
  Ring.fromBigInt[A](bi) ?== Ring.defaultFromBigInt[A](bi)
}
```
you have a kind of homomorphism here, that seems like a stronger law, isn't it?

tixxit: Exactly - we do have a homomorphism here, but I don't think the law is too strong. Any ring has a homomorphism from Z -> R (BigInt -> A) and this is really just codifying this.

johnynek: okay, as long as this is not one of those "it exists but finding the homomorphism is not always trivial" type things, I'm +1.

tixxit: Agreed and this is not that - you can always use `defaultFromBigInt` as the implementation, which should pass the laws (assuming `fromInt` does too)!

non:
This looks good to me. 👍

PR #174, opened 2016-10-05
Corrects EuclideanRing instances.

denisrosset:
The quotmod operation is uniquely defined in fields; this PR provides the correct implementation for Field.

This fixes the BigDecimal, Float, Double instances so that they are lawful (the test Rat instance was correct).

However, as we do not implement `euclideanFunction` in algebra, and the BigDecimal, Float and Double are partially tested, the incorrect behavior did not fail any tests --- and introducing failing tests requires quite a bit of infrastructure (see #172).

Note: there is a substantial impact on Spire as the operator `%` correspond to `EuclideanRing.mod`. Before the change, the incorrect instance gives `3.14 % 2 = 0.14`; this behavior is consistent with the scala stdlib syntax. After the change, `3.14 % 2 = 0`. Thus, `%` should be implemented using another type (see `tquotmod` in #172).

Note 2: the proof that `EuclideanRing.mod(x,y)=0` in a Field can be found in standard algebra textbooks. However, the jargon is quite hermetic. I can adapt an elementary proof from the law axioms if really needed.

### DIFF
--- a/core/src/main/scala/algebra/instances/bigDecimal.scala
+++ b/core/src/main/scala/algebra/instances/bigDecimal.scala
@@ -21,13 +21,6 @@ class BigDecimalAlgebra extends Field[BigDecimal] with Serializable {
 
   def times(a: BigDecimal, b: BigDecimal): BigDecimal = a * b
   def div(a: BigDecimal, b: BigDecimal): BigDecimal = a / b
-  def quot(a: BigDecimal, b: BigDecimal) = a.quot(b)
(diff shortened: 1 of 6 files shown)


johnynek:
I'm +1 on this.

denisrosset:
Alternative: remove EuclideanRing from algebra, and provide a complete hierarchy of commutative ring types in spire.

non:
As far as algebra goes 👍 

I'm not totally sure how things will shake out in Spire but this works for me.

PR #181, opened 2016-11-24
Remove EuclideanRing.

tixxit:
This removes `EuclideanRing` from algebra. `EuclideanRing` is only used by Spire, so Spire will keep `EuclideanRing`, and Spire's `Field` will extend both its own `EuclideanRing` and algebra's `Field`.


core/src/main/scala/algebra/instances/long.scala
```diff
@@ -7,14 +7,14 @@ import algebra.ring._
 package object long extends LongInstances
 
 trait LongInstances extends cats.kernel.instances.LongInstances {
-  implicit val longAlgebra: EuclideanRing[Long] =
+  implicit val longAlgebra: LongAlgebra =
```
johnynek: why not `CommutativeRing[Long]`? Is the idea that the binary compat is more stable if we point to a single stable class?

tixxit: So, there are 2, possibly terrible, reasons:

The first is that if you use machinist for ops, then when you say something like: `x + y`, it gets translated into `long.longAlgebra.plus(x, y)`. If `longAlgebra`'s return type is a class, then we get an `invokevirtual` to call `plus`. If `longAlgebra` is an interface, we get an `invokeinterface`. The JVM can better optimize the former and there are some minor performance improvements.

2. It fixes future possible mistakes. If we just say it returns `LongAlgebra`, then if someone adds a new `with SomeFutureTypeclass[Long]` to `LongAlgebra`, then we don't have to worry about them forgetting to add it to the return type of the implicit too.

johnynek: Okay. Seems legit to me, I guess.

johnynek:
:+1:

non:
👍

PR #182, opened 2016-11-24
Remove code duplicated from cats-kernel.

tixxit:
Much of our laws code has been moved into cats-kernel, so this just gets rid of it and uses the stuff from cats-kernel. It's also a win, since cats-kernel has since added more tests!


tixxit:
I also changed the RingLaws a bit so we only use the non-zero predicate for the actual group laws, but still allow zeros in, eg, the multiplicative monoid laws.

laws/src/test/scala/algebra/laws/Rat.scala
```diff
@@ -99,9 +99,17 @@ object Rat {
   val RatMinMaxLattice: DistributiveLattice[Rat] =
     DistributiveLattice.minMax[Rat](ratAlgebra)
 
-  implicit val ratArbitrary =
+  // Is this horrible? Yes. Am I ashamed? Yes.
+  private[this] def genNonZero: Gen[BigInt] =
+    arbitrary[BigInt].flatMap { x =>
```
johnynek: need tailRecM here! ;)

johnynek:
:+1:

non:
👍

PR #184, opened 2016-11-25
Scala 2.12: updated to cats 0.8.1 and shadowed its dependencies.

denisrosset:
Everything is basically in the title!

...except the `Order` that depends on the generation of `A => A`, which should be updated to use a `Cogen[A]`.


tixxit:
@denisrosset If we merge #182 first, then we won't have to worry about the `Order` test.

thomas-stripe:
@denisrosset OK, #182 got merged, so we just use cats-kernel's `OrderLaws`. Want to merge master?

denisrosset:
Merged and bug fixed.

johnynek:
👍

PR #190, opened 2016-11-30
Split Field.fromDouble to enable reuse.

denisrosset:
e.g. in Spire DivisionRing. The code is quite heavy in bit tricks that are difficult to verify, so I'd like to avoid duplication.

### DIFF
--- a/core/src/main/scala/algebra/ring/Ring.scala
+++ b/core/src/main/scala/algebra/ring/Ring.scala
@@ -74,4 +74,39 @@ trait RingFunctions[R[T] <: Ring[T]] extends AdditiveGroupFunctions[R] with Mult
 
 object Ring extends RingFunctions[Ring] {
   @inline final def apply[A](implicit ev: Ring[A]): Ring[A] = ev
+
+  /** Returns the given Double, understood as a rational number, in the provided
+    * (division) ring.
+    * 
+    * This is implemented in terms of basic ops. However, this is
+    * probably significantly less efficient than can be done with a specific
+    * type. So, it is recommended to specialize this general method.
+    */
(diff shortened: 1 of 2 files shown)


core/src/main/scala/algebra/ring/Ring.scala
```diff
@@ -74,4 +74,39 @@ trait RingFunctions[R[T] <: Ring[T]] extends AdditiveGroupFunctions[R] with Mult
...
   @inline final def apply[A](implicit ev: Ring[A]): Ring[A] = ev
+
+  /** Returns the given Double, understood as a rational number, in the provided
+    * (division) ring.
+    * 
+    * This is implemented in terms of basic ops. However, this is
+    * probably significantly less efficient than can be done with a specific
+    * type. So, it is recommended to specialize this general method.
+    */
+  def fromDouble[A](a: Double)(implicit ringA: Ring[A], mgA: MultiplicativeGroup[A]): A =
```
tixxit: Not sure I like this in `Ring`, since it can't actually be used with only a `Ring`. I think we should rename this to `defaultFromDouble` too, since we want to encourage folks to use the `fromDouble` on, eg, `Field`, which may be overridden to be faster.

tixxit: I just noticed this was in algebra, not spire! So, yeah, I guess Ring isn't too bad here. But I still think we should rename this `defaultFromDouble`.

johnynek:
:+1:

non:
Seems reasonable to me. 👍

non:
@denisrosset Do you want to add a test that calls this method? Or should we merge and worry about coverage later?

tixxit:
@non Yeah, that PR will exercise this code through `Rat`, which uses the default implementation of `fromDouble`.

PR #191, opened 2016-11-30
Add test for Field.fromDouble.

tixxit:
This adds a test for `fromDouble` - it uses the more brute-force version where we convert the Double to a BigDecimal, use that to extract a numerator/denominator, then use `Ring.fromBigInt` on both and divide.


johnynek:
:+1: nice bump in coverage!

non:
👍

PR #192, opened 2016-12-01
Test field laws for Float/Double/BigDecimal.

tixxit:
This tests the `Field` instances for `Float`, `Double` and `BigDecimal` in a bit of a sneaky way. Rather than test them directly, it uses a `FPApprox` wrapper type. This exercises the underlying `Field` to handle the regular operations, but also maintains an error bound on the approximation. The `Eq` instance for `FPApprox[A]` will return true for any 2 numbers that could be equal, given their error bounds. Of course, there are more subtleties than that:

The numbers occasionally underflow, which causes all sorts of annoying problems. So, we also have to occasionally check if values after multiplications are near-zero and, if so, bump them back up to a very small, but useable number.

The numbers occasionally overflow to infinity, at which point the game is lost, since we soon start getting NaNs. So, we also return `true` if a comparison gets us into non-finite number territory.

### DIFF
--- a/core/src/main/scala/algebra/instances/bigDecimal.scala
+++ b/core/src/main/scala/algebra/instances/bigDecimal.scala
@@ -1,19 +1,20 @@
 package algebra
 package instances
 
+import java.math.MathContext
+
 import algebra.ring._
 
 package object bigDecimal extends BigDecimalInstances
 
 trait BigDecimalInstances extends cats.kernel.instances.BigDecimalInstances {
   implicit val bigDecimalAlgebra: BigDecimalAlgebra =
-    new BigDecimalAlgebra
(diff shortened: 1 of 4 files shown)


tixxit:
For whatever reason, the `fromDouble` test is failing for `FPApprox[Float]` and `FPApprox[Double]` on JS only. The values I'm getting make no sense and it seems suspiciously like a bug in `BigInt`. Need to investigate further...

tixxit:
Ugh - so, this is because the `fromDouble` test uses `BigDecimal`, which is apparently broken on scalajs. I'm going to skip the tests if `Platform.isJVM` is false, which should be OK for now. If BigDecimal gets unbroken, we can re-enable.

erik-stripe:
👍

PR #194, opened 2016-12-03
Add JMH-based benchmark project with example benchmarks

### LINKED ISSUE #31: Add benchmarking project
It would be good to have a benchmarking project defined. This would allow us to do a few things:
1. Get benchmarks for `std` instances.
2. Confirm that miniboxing/specialization is working correctly.
3. Avoid performance regressions for helper methods.

What do you all think? Spire has done this and it's been really useful over the years.

sritchie:
Closes #31. This PR also

- Updates binary compatibility information
- Removes the MIMA check from the build because we've already broken binary compatibility with 0.6.0!!
- fixes some busted info in `publishSettings`


johnynek:
How can we have broken mima compat with 0.6.0 if mima is turned on? If we broke compatibility we need to fix it. Or make a new release with mima on.

We are going to get burned if we have this at the bottom of algebird and Spire but break compatibility all the time.

sritchie:
None of the projects were specifying their youngest compatible versions. I
added those and boom, failure.

non:
Oh jeez. Can you add a comment with the binary incompatibilities?

sritchie:
yeah, one second...

sritchie:
```
[info] core: found 1 potential binary incompatibilities while checking against org.typelevel:algebra_2.12:0.6.0
[error]  * method this()Unit in class algebra.instances.BigDecimalAlgebra does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("algebra.instances.BigDecimalAlgebra.this")
[info] Done updating.
[info] benchmark: previous-artifact not set, not analyzing binary compatibility
[info] laws: found 0 potential binary incompatibilities while checking against org.typelevel:algebra-laws_2.12:0.6.0
[info] laws: found 0 potential binary incompatibilities while checking against org.typelevel:algebra-laws_2.12:0.6.0
[info] core: found 1 potential binary incompatibilities while checking against org.typelevel:algebra_2.12:0.6.0
[error]  * method this()Unit in class algebra.instances.BigDecimalAlgebra does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("algebra.instances.BigDecimalAlgebra.this")
```

johnynek:
Ahh, this was Thomas's change to accept a parameter to the constructor. We
can fix that by adding a default constructor with no params. Problem solved.

def this() = this(BigDecimal...

non:
👍 Thanks for catching this @sritchie

sritchie:
@johnynek @non that addition worked great! I changed it and re-enabled mima. let's see how the build goes.

non:
If that works I can try to do a rush 0.6.1 release to preserve 0.5.x compatibility.

sritchie:
Ah, is this 0.5 compatible too? If so feel free to change the minimum
version in this PR! I've gotta pull myself away for the weekend.

non:
No, sorry, I got confused on which version we were on. Ignore me!

sritchie:
Hmmm, looks like we have another legit binary compatibility issue. This commit is the culprit: https://github.com/typelevel/algebra/commit/24fe85a71d58056d07c954f9a1a900159a053ea8

it renames `Ring.fromDouble` to `defaultFromDouble`.
```
[error]  * method defaultFromDouble(Double,algebra.ring.Ring,algebra.ring.MultiplicativeGroup)java.lang.Object in trait algebra.ring.RingFunctions is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("algebra.ring.RingFunctions.defaultFromDouble")
[info] core: found 1 potential binary incompatibilities while checking against org.typelevel:algebra_2.11:0.6.0
[error]  * method defaultFromDouble(Double,algebra.ring.Ring,algebra.ring.MultiplicativeGroup)java.lang.Object in trait algebra.ring.RingFunctions is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("algebra.ring.RingFunctions.defaultFromDouble")
```

@non @johnynek, do one of you want to fix this in a separate PR, or commit to this PR? Might be better to use this one to leverage the mima check.

johnynek:
Maybe it is best to make a PR to turn Mima back on and fix the issues (or quickly make a 0.7.0 of algebra).

One idea to fix issues would be to add methods to `RingFunctions[R]` using an `implicit class`. Kind of round about, but safe for binary compatibility and makes source compatibility closer to binary compatibility.

sritchie:
:boom: done @johnynek, and thanks for the suggested fix!

thomas-stripe:
👍

sritchie:
@thomas-stripe, can I trouble you for a merge?

PR #213, opened 2018-02-12
Update to Cats 1.0.1

vpavkin:
My attempt to continue progress on #207.

@johnynek is this all we need ? :)

PR #214, opened 2018-02-20
Released 1.0.0

coltfred:
I released 1.0.0  cc/ @johnynek 

I don't have the ability to push to the actual project so these are the changes that were made as part of that.

#213 was the reason I did the publish.


johnynek:
thanks a ton @coltfred !

Big help.

PR #215, opened 2018-02-20
Update README to reference 1.0.0 as latest version

coltfred:


johnynek:
👍

johnynek:
well merge when green (make sure we don't have any CI checks that could fail on README, we should have them...) We could have a CI check to see if the version is actually the latest on maven-central. That would be cool.

PR #222, opened 2019-01-30
fix installation of gems

larsrh:


larsrh:
I'm fast-tracking this without review to fix the build.

PR #223, opened 2019-01-30
build for 2.13.0-M5

larsrh:
- simplify build (depend on cats-testkit directly, simplify dependencies)
- drop 2.10.x build (dropped by cats)
- bump dependencies

MiMa fails the check against 0.6.0, but succeeds against 1.0.0. So, if nobody complains, I'd call the next release with this PR merged 1.0.1.


erikerlandson:
my understanding was that this entire project was going to be moved into a new `cats-algebra` subpackage

larsrh:
Apparently that's typelevel/cats#2041. Still, I'd like to get a 2.13.0-M5 compatible release out. Who knows when the merge is going to happen.

erikerlandson:
I tried this via #221, but either way, could also make use of a 2.13 build

larsrh:
I didn't see that PR, unfortunately :frowning_face:

PR #224, opened 2019-03-29
De Morgan algebra

vquintin:
* Add `DeMorgan` typeclass for fuzzy logic

* `Bool` extends `DeMorgan`

* Extract trait `Logic` common to both `DeMorgan` and `Heyting`

* Add laws for `DeMorgan`

The motivation of this PR is to fix the invalid `Heyting` instance for `Trilean` in [spire](https://github.com/typelevel/spire/blob/master/core/src/main/scala/spire/math/Trilean.scala). Trilean is a `DeMorgan`, not a `Heyting` (it does not satisfy the law of non contradiction).

### DIFF
--- a/core/src/main/scala/algebra/lattice/Bool.scala
+++ b/core/src/main/scala/algebra/lattice/Bool.scala
@@ -23,8 +23,7 @@ import scala.{specialized => sp}
  * Every boolean algebras has a dual algebra, which involves reversing
  * true/false as well as and/or.
  */
-trait Bool[@sp(Int, Long) A] extends Any with Heyting[A] with GenBool[A] { self =>
-  def imp(a: A, b: A): A = or(complement(a), b)
+trait Bool[@sp(Int, Long) A] extends Any with DeMorgan[A] with Heyting[A] with GenBool[A] { self =>
   def without(a: A, b: A): A = and(a, complement(b))
 
   // xor is already defined in both Heyting and GenBool.
--- a/core/src/main/scala/algebra/lattice/Bool.scala
+++ b/core/src/main/scala/algebra/lattice/Bool.scala
@@ -23,7 +23,8 @@ import scala.{specialized => sp}
  * Every boolean algebras has a dual algebra, which involves reversing
  * true/false as well as and/or.
  */
-trait Bool[@sp(Int, Long) A] extends Any with DeMorgan[A] with Heyting[A] with GenBool[A] { self =>
+trait Bool[@sp(Int, Long) A] extends Any with Heyting[A] with GenBool[A] { self =>
+  def imp(a: A, b: A): A = or(complement(a), b)
   def without(a: A, b: A): A = and(a, complement(b))
 
   // xor is already defined in both Heyting and GenBool.
--- a/core/src/main/scala/algebra/lattice/DeMorgan.scala
+++ b/core/src/main/scala/algebra/lattice/DeMorgan.scala
@@ -5,15 +5,14 @@ import scala.{specialized => sp}
 
 /**
   * De Morgan algebras are bounded lattices that are also equipped with
-  * an involutive `complement` operation.
+  * a De Morgan involution.
   *
-  * Complementation obeys the following law:
+  * De Morgan involution obeys the following laws:
   *
   *  - ¬¬a = a
   *  - ¬(x∧y) = ¬x∨¬y
   *
-  * However, in De Morgan algebras this operation is only a
-  * pseudo-complement, since De Morgan algebras do not necessarily
+  * However, in De Morgan algebras this involution does not necessarily
   * provide the law of the excluded middle. This means that there is no
   * guarantee that (a ∨ ¬a) = 1. De Morgan algebra do not not necessarily
   * provide the law of non contradiction either. This means that there is
@@ -21,7 +20,7 @@ import scala.{specialized => sp}
   *
   * De Morgan algebras are useful to model fuzzy logic. For a model of
   * classical logic, see the boolean algebra type class implemented as
-  * `Bool`.
+  * [[Bool]].
   */
 trait DeMorgan[@sp(Int, Long) A] extends Any with BoundedDistributiveLattice[A] with Logic[A] { self =>
   def meet(a: A, b: A): A = and(a, b)
--- a/core/src/main/scala/algebra/lattice/DeMorgan.scala
+++ b/core/src/main/scala/algebra/lattice/DeMorgan.scala
@@ -51,4 +51,22 @@ object DeMorgan extends DeMorganFunctions[DeMorgan] with DeMorganGenBoolOverlap[
     * Access an implicit `DeMorgan[A]`.
     */
   @inline final def apply[@sp(Int, Long) A](implicit ev: DeMorgan[A]): DeMorgan[A] = ev
+
+
+  /**
+    * Turn a [[Bool]] into a `DeMorgan`
+    * Used for binary compatibility.
+    */
+  final def fromBool[@sp(Int, Long) A](bool: Bool[A]): DeMorgan[A] =
+    new DeMorgan[A] {
+      def and(a: A, b: A): A = bool.and(a, b)
+
+      def or(a: A, b: A): A = bool.or(a, b)
+
+      def complement(a: A): A = bool.complement(a)
+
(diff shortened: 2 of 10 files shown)


larsrh:
This would be a binary-breaking change and requires a 1.x release.

Cats team, where are we with the incorporation of algebra into cats? @kailuowang @lukajcb

core/src/main/scala/algebra/lattice/DeMorgan.scala
```diff
@@ -0,0 +1,55 @@
+package algebra
+package lattice
+
+import scala.{specialized => sp}
+
+/**
+  * De Morgan algebras are bounded lattices that are also equipped with
+  * an involutive `complement` operation.
```
TomasMikula: If it is not necessarily a _complement_, let's not call it that. Just call it _De Morgan involution_ instead?

core/src/main/scala/algebra/lattice/DeMorgan.scala
```diff
@@ -0,0 +1,55 @@
...
+  *
+  * Complementation obeys the following law:
+  *
+  *  - ¬¬a = a
+  *  - ¬(x∧y) = ¬x∨¬y
+  *
+  * However, in De Morgan algebras this operation is only a
+  * pseudo-complement, since De Morgan algebras do not necessarily
```
TomasMikula: The definition of pseudocomplement requires `(a ∧ ¬a) = 0`, which, as you mention below, does not necessarily hold for De Morgan algebras.

core/src/main/scala/algebra/lattice/DeMorgan.scala
```diff
@@ -0,0 +1,55 @@
...
+  * provide the law of the excluded middle. This means that there is no
+  * guarantee that (a ∨ ¬a) = 1. De Morgan algebra do not not necessarily
+  * provide the law of non contradiction either. This means that there is
+  * no guarantee that (a ∧ ¬a) = 0.
+  *
+  * De Morgan algebras are useful to model fuzzy logic. For a model of
+  * classical logic, see the boolean algebra type class implemented as
+  * `Bool`.
```
TomasMikula: Let's use a Scaladoc link:

```scala
/**
  * For a model of classical logic, see [[Bool]].
  */
```

core/src/main/scala/algebra/lattice/Heyting.scala
```diff
@@ -33,20 +33,10 @@ import scala.{specialized => sp}
  * classical logic, see the boolean algebra type class implemented as
  * `Bool`.
  */
-trait Heyting[@sp(Int, Long) A] extends Any with BoundedDistributiveLattice[A] { self =>
-  def and(a: A, b: A): A
+trait Heyting[@sp(Int, Long) A] extends Any with BoundedDistributiveLattice[A] with Logic[A] { self =>
```
johnynek: similar comment about not breaking compatibility here. Can we make `Heyting` from `Logic`?

core/src/main/scala/algebra/lattice/Logic.scala
```diff
@@ -0,0 +1,55 @@
...
+package lattice
+
+import scala.{specialized => sp}
+
+/**
+  * Logic models a logic generally.
+  * For intuitionistic logic see `Heyting`
+  * For fuzzy logic see `DeMorgan`
+  */
+trait Logic[@sp(Int, Long) A] extends Any { self =>
```
TomasMikula: Is this typeclass lawless?

core/src/main/scala/algebra/lattice/DeMorgan.scala
vquintin: Yes, De Morgan involution is a better name. I will change it :+1:

core/src/main/scala/algebra/lattice/Logic.scala
vquintin: It is lawless at the moment. I am not sure what general laws common to both `Heyting` and `DeMorgan` I could use here.
If you have any input, I am interested to read it.

TomasMikula: They are both _bounded distributive lattices_.

One common property of the De Morgan involution and the pseudocomplement operation is that _a ≤ ¬¬a_. Not sure if it is terribly useful.

The question here is, what use of `Logic` do you imagine?

vquintin: I want to express a structure which is very much like a *bounded distributive lattice* with an extra unary operator.
I would like it not to use lattice operations (`meet, join`) but use instead common logic operators (`and, or, xor, not, imp,...`).

Do you think I should skip the `Logic` bit and come up with a name for a *bounded distributed lattice* with an extra unary operator ?
i.e. in the company code I am currently making open-source, `DeMorgan` is actually `InvolutiveBoundedLattice`

TomasMikula: > I want to express a structure which is very much like a bounded distributive lattice with an extra unary operator.

Surely you don't want any arbitrary unary operation. So let's try to formulate what properties it should satisfy in relation to the other operations.

> Do you think I should skip the `Logic` bit and come up with a name for a bounded distributed lattice with an extra unary operator ?

I am not concerned about the name, yet.

From [Wikipedia](https://en.wikipedia.org/wiki/Heyting_algebra#The_De_Morgan_laws_in_a_Heyting_algebra):

> One of the two De Morgan laws is satisfied in every Heyting algebra, namely
>
>     ¬(x ∨ y) = ¬x ∧ ¬y
>
> However, the other De Morgan law does not always hold. We have instead a weak de Morgan law:
>
>     ¬(x ∧ y) = ¬¬(¬x ∨ ¬y)

kailuowang:
@larsrh sorry for the late reply. I think we are all on board in terms of moving it in. If anyone has time to make the PR I think we can make it happen in the next release, or the one after.

core/src/main/scala/algebra/lattice/Bool.scala
```diff
@@ -23,8 +23,7 @@ import scala.{specialized => sp}
  * Every boolean algebras has a dual algebra, which involves reversing
  * true/false as well as and/or.
  */
-trait Bool[@sp(Int, Long) A] extends Any with Heyting[A] with GenBool[A] { self =>
-  def imp(a: A, b: A): A = or(complement(a), b)
+trait Bool[@sp(Int, Long) A] extends Any with DeMorgan[A] with Heyting[A] with GenBool[A] { self =>
```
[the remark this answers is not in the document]
vquintin: Thanks for your help on binary compatibility ! I will fix it.

larsrh:
@kailuowang OK, so I'd say we merge this in here regardless. Porting to Cats can happen at a later time.

LukaJCB:
In terms of merging Cats, I think we still haven't decided how exactly we want to do that, if we want to have a new module vs. merging it into `cats-core` or `cats-kernel`. Though maybe it is best to just have a new module and then atleast have it in the same repo.

larsrh:
Blocked by #225. If we want this to be binary compatible, we should enforce it.

core/src/main/scala/algebra/lattice/Logic.scala
```diff
@@ -0,0 +1,79 @@
...
+  def not(a: A): A
+
+  def xor(a: A, b: A): A = or(and(a, not(b)), and(not(a), b))
+  def nand(a: A, b: A): A = not(and(a, b))
+  def nor(a: A, b: A): A = not(or(a, b))
+  def nxor(a: A, b: A): A = not(xor(a, b))
+}
+
+trait LogicGenBoolOverlap[H[A] <: Logic[A]] {
+  def and[@sp(Int, Long) A](x: A, y: A)(implicit ev: H[A]): A =
```
TomasMikula: What is the point of this trait, if it is only extended once (by `object Logic`, but not by `object GenBool`? Same goes for `DeMorganGenboolOverlap`.

larsrh:
@vquintin I've merged in #224 and pushed to your branch to enable MiMa checking.

larsrh:
@johnynek Are we good to merge this?

larsrh:
1.0.2 release containing this change and 2.13.0-RC1 compatibility forthcoming.

PR #226, opened 2019-04-25
[WIP] Scala 2.13.0-RC1

larsrh:
Tests don't compile currently.

Could use some help by @kailuowang; probably caused by the Cats upgrade.


kailuowang:
build errors on 2.12- seem to be caused by the fact that cats removed catalyst dependency. algebra can no longer depend on `Catalyst.platform`.  It may need to copy the Platform utility code like Cats did
https://github.com/typelevel/cats/blob/master/kernel-laws/js/src/main/scala/cats/platform/Platform.scala

The build error on 2.13 seems be caused by .travis referencing the wrong scala 2.13 version (2.13.0-M4 vs. 2.13.0-RC1)

larsrh:
Why not export that object?

kailuowang:
It's so trivial and cats doesn't seem to be the right place for it to live in. A sbt-plugin might be more suitable.

larsrh:
OK, I will whip up something based on sbt-buildinfo.

larsrh:
One more issue I don't understand:

```
[error] /home/travis/build/typelevel/algebra/core/src/test/scala/algebra/ring/RingTest.scala:10:12: could not find implicit value for parameter genA: org.scalatest.prop.Generator[BigInt]
[error]     forAll { (n: BigInt) =>
[error]            ^
[error] one error found
```

Why are we suddenly looking for a ScalaTest gen and not a ScalaCheck one?

kailuowang:
@larsrh his is a major breaking change from scalatest 3.0.x to 3.1.0. 
3.1 introduced its own prop testing functionality. To continue use ScalaCheck, your test suite needs to extend `ScalaCheckPropertyChecks`(which is provided by `scalatestplus-scalacheck` instead of `GeneratorDrivenPropertyChecks` 

You can see my changes to Cats here
https://github.com/typelevel/cats/pull/2806/files

Unfortunately there is no migration guide for scalatest 3.1.0 yet, so I had to read their code and a bit commit history. It looks like that someone, if not us, has to write up a migration guide sooner rather than later.

larsrh:
Progress: portable-scala/sbt-crossproject#108

larsrh:
Closed in favour of #228.

PR #228, opened 2019-05-17
Updates for 2.13.0-RC1

travisbrown:
This includes @larsrh's commit from #226 and does a few remaining things needed for 2.13, including removing Catalysts (by copy-pasting the platform-checking stuff and scrapping the macro-powered type tag in favor of explicit type names for laws).

I've also gone ahead and updated a few of the sbt plugins and the older Scala versions.

MiMa says it's fine. The new `algebra.laws.platform.Platform` thing is only Scala-private (you could use it from Java if you really wanted to), but that doesn't seem like a big deal.

I'm doing this because I'd like to have Spire on 2.13. I guess it could wait until everything upstream is out for 2.13.0-RC2, but given that that includes ScalaTest and Cats it could take a while.

r? @larsrh


larsrh:
Thanks a lot! Seems like bumps for `tut` and `mdoc` are still required.

travisbrown:
@larsrh mdoc isn't published for any 2.13 versions, so I think it makes sense just to disable the `makeMicrosite` check in CI (this is what I've done in my other projects). The 2.13.0-RC1 build is also failing because MiMa has no previous artifacts to compare to, so I've also disabled that check. When this is merged I'll open issues to remind us to re-enable these checks when we can.

larsrh:
Can I get another :+1:? @johnynek?

PR #236, opened 2020-04-15
Update sbt and Scala

jhnsmth:


jhnsmth:
seems like this project is missing @scala-steward

larsrh:
This repository is in maintenance-only mode because of cats/issues#2041, although the reality might be that nobody is picking up the slack (it's menial work).

larsrh:
Merging this despite maintenance mode. Thanks for your contribution!

PR #239, opened 2020-12-10
Dotty

cquiroz:
This PR contains a bunch of updates to libraries including:

* Update to cats 2.3.0
* Update scala.js
* Port tests to munit
* Cross compile to dotty

There is a mima check failure but I think it comes from cats. I'm not too familiar with mima, should an exception be made?


build.sbt
```diff
@@ -155,12 +157,15 @@ lazy val core = crossProject(JSPlatform, JVMPlatform)
   .settings(moduleName := "algebra")
   .settings(
     mimaPreviousArtifacts := Set("org.typelevel" %% "algebra" % binaryCompatibleVersion),
-    mimaBinaryIssueFilters ++= ignoredABIProblems
```
johnynek, line 158: Why did you remove this?

laws/shared/src/test/scala/algebra/laws/Rat.scala
```diff
@@ -47,14 +47,14 @@ class Rat(val num: BigInt, val den: BigInt) extends Serializable { lhs =>
     if (num == 0) throw new ArithmeticException("/0") else Rat(den, num)
 
   def /(rhs: Rat): Rat =
-    lhs * rhs.reciprocal
+    lhs * rhs.reciprocal()
 
   def **(k: Int): Rat =
     Rat(num.pow(k), den.pow(k))
 
   def toDouble(): Double = num.toDouble / den.toDouble
```
johnynek, line 55: Style wise, these pure functions shouldn't use () right?

laws/shared/src/test/scala/algebra/laws/Rat.scala
```diff
@@ -123,14 +123,14 @@ class RatAlgebra extends Field[Rat] with Order[Rat] with Serializable {
   def plus(a: Rat, b: Rat): Rat = a + b
   def negate(a: Rat): Rat = -a
   def times(a: Rat, b: Rat): Rat = a * b
-  override def reciprocal(a: Rat): Rat = a.reciprocal
+  override def reciprocal(a: Rat): Rat = a.reciprocal()
```
johnynek, line 126: Shouldn't Rat not use () on reciprocal?

johnynek (commented):
Thanks so much for taking this!

I'm very excited to see dotty be supported.

johnynek:
I think the mima errors are that you removed the ignoredABI thing from mima in the build.

I think if you put that back it will be okay.

laws/shared/src/test/scala/algebra/laws/Rat.scala
cquiroz: Right, it maybe make sens to remove the parentheses, I'm not sure if it will break mima

build.sbt
johnynek: Removing unneeded parens shouldn't change mima. The JVM binary generated should be the same.

cquiroz: Actually it wasn't removed it was moved down,the diff doesn't make it very clear

cquiroz:
I suspect the mima error is due to:
https://github.com/typelevel/cats/commit/0f2215111a2607469ace291ecd86f8fb05730035#diff-1486a9963346d9a4200ebe5d66128d4904963a678c937ef8f9559ff5ced08972R5

The signature of ` catsKernelStdOrderForChar` changed

johnynek (approved):
Thanks!

cquiroz:
It would be great to get a release. I'd like to try to get `spire` to cross build to dotty

larsrh:
I hadn't noticed this PR. I will get a release done, thanks!

PR #241, opened 2021-02-13
Enable Scala Native build for all published modules.

arashi01:
Enables Scala Native build for published modules.

- Migrate from deprecated Tut sbt plugin to Mdoc plugin. (Required for upgrade to Scala 2.12.13)
- Migrate from deprecated actions/install-ruby github action to ruby/install-ruby action


arashi01:
I think we're ready to go on this. Last commit, splitting tests per platform is optional. Can revert if original setup is preferred.

cc @larsrh

larsrh:
Excellent, thanks!

larsrh:
2.2.0 is out

PR #243, opened 2021-02-18
Scala 3.0.0-RC1

larsrh:


core/src/main/scala/algebra/lattice/Heyting.scala
```diff
@@ -59,8 +59,8 @@ trait HeytingGenBoolOverlap[H[A] <: Heyting[A]] {
 }
 
 trait HeytingFunctions[H[A] <: Heyting[A]] extends
-  BoundedMeetSemilatticeFunctions[H] with
-  BoundedJoinSemilatticeFunctions[H] {
+  BoundedMeetSemilatticeFunctions[H]
+  with BoundedJoinSemilatticeFunctions[H] {
```
johnynek, line 63: these all just seem like whitespace changes. Am I missing something? Does this really break 3.0-RC1? If so, this should be release blocker IMO.

larsrh: It does break RC1 compilation, yes.

johnynek: This is a bug, right?

larsrh: Not sure. I'm going to find out.

johnynek: It seems to compile the code, but loses track of the fact that the mixin was there.... That seems like a fairly serious bug to me.

larsrh: No, there's an actual syntax error there.

larsrh: I've been told that this behaviour change was intentional.

johnynek: so, it is a syntax error now to have a trailing with even without braceless syntax?

mpilquist: This was a bug in RC1 which was just fixed - looks like it was missed when `with`-as-template-start was removed in favor of `:`: https://github.com/lampepfl/dotty/pull/11448

johnynek: I'm so confused not the braceless syntax. I thought `with` was back to being the way to go... Is it colons again or something else? That PR would be nicer if it linked to some documentation of the latest version of syntax.

mpilquist: For braceless syntax, a trailing colon creates a lexical scope, like:

```scala
class Foo:
  def foo: Int =
    val x = 1
    val y = 2
    x + y

object Foo:
  def apply: Foo = new Foo
```

More examples in the official docs here: https://dotty.epfl.ch/docs/reference/other-new-features/indentation.html

After M3 and before RC1, there was an experiment where `:` was replaced with `with` -- that didn't work out and was rolled back but part of the rollback was missed, and hence the build failure we have here. 11448 fixes that and will be in RC2. Hope that helps.

PR #245, opened 2021-04-27
Update sbt and scala

cquiroz:
Mostly we get scala 3.0.0-RC3 but also the build is improved


.github/workflows/ci.yml
```diff
@@ -23,14 +23,14 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        scala: [2.12.13, 2.13.4, 3.0.0-RC1, 3.0.0-RC2]
+        scala: [2.12.13, 2.13.5, 3.0.0-RC2, 3.0.0-RC3]
```
johnynek, line 26: do we need RC2 still? I think we aren't changing this repo much, so I would imagine just doing RC3 would be fine.

johnynek:
Thanks!

.github/workflows/ci.yml
cquiroz: Probably not, I'm just keeping this idea of releasing the last 2 versions

cquiroz:
Would it be possible to make a release?

PR #246, opened 2021-05-14
Adding GCD rings and Euclidean rings, along with instances for BigInt

denisrosset:
This aligns the commutative rings/field hierarchy with the one available in Spire.

I haven't provided instances for non-exact types such as Long, Int, etc, because their overflow behaviour doesn't work with the laws.

(Spire tests such instances by aborting the test in case of an overflow -- which is then considered undefined behaviour).

### DIFF
--- a/core/src/main/scala/algebra/ring/DivisionRing.scala
+++ b/core/src/main/scala/algebra/ring/DivisionRing.scala
@@ -1,0 +1,22 @@
+package algebra
+package ring
+
+import scala.{specialized => sp}
+
+trait DivisionRing[@sp(Byte, Short, Int, Long, Float, Double) A] extends Any with Ring[A] with MultiplicativeGroup[A] {
+  self =>
+
+  def fromDouble(a: Double): A = Field.defaultFromDouble[A](a)(self, self)
+
+}
+
+trait DivisionRingFunctions[F[T] <: DivisionRing[T]] extends RingFunctions[F] with MultiplicativeGroupFunctions[F] {
+  def fromDouble[@sp(Int, Long, Float, Double) A](n: Double)(implicit ev: F[A]): A =
+    ev.fromDouble(n)
+}
+
+object DivisionRing extends DivisionRingFunctions[DivisionRing] {
+
+  @inline final def apply[A](implicit f: DivisionRing[A]): DivisionRing[A] = f
+
+}
--- a/core/src/main/scala/algebra/ring/EuclideanRing.scala
+++ b/core/src/main/scala/algebra/ring/EuclideanRing.scala
@@ -1,0 +1,59 @@
+package algebra
+package ring
+
+import scala.annotation.tailrec
+import scala.{specialized => sp}
+
+/**
+ * EuclideanRing implements a Euclidean domain.
+ *
+ * The formal definition says that every euclidean domain A has (at
+ * least one) euclidean function f: A -> N (the natural numbers) where:
+ *
+ *   (for every x and non-zero y) x = yq + r, and r = 0 or f(r) < f(y).
+ *
+ * This generalizes the Euclidean division of integers, where f represents
+ * a measure of length (or absolute value), and the previous equation
+ * represents finding the quotient and remainder of x and y. So:
+ *
+ *   quot(x, y) = q
+ *   mod(x, y) = r
+ */
+trait EuclideanRing[@sp(Int, Long, Float, Double) A] extends Any with GCDRing[A] { self =>
+  def euclideanFunction(a: A): BigInt
+  def equot(a: A, b: A): A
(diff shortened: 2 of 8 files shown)


core/src/main/scala/algebra/ring/EuclideanRing.scala
```diff
@@ -0,0 +1,59 @@
...
+  @inline final def apply[A](implicit e: EuclideanRing[A]): EuclideanRing[A] = e
+
+  /**
+   * Simple implementation of Euclid's algorithm for gcd
+   */
+  @tailrec final def euclid[@sp(Int, Long, Float, Double) A: Eq: EuclideanRing](a: A, b: A): A = {
+    if (EuclideanRing[A].isZero(b)) a else euclid(b, EuclideanRing[A].emod(a, b))
+  }
+
+/*  @tailrec final def extendedEuclid[@sp(Int, Long, Float, Double) A: Eq: EuclideanRing](a: A, b: A): (A, A, A) = {
```
johnynek, line 56: Can we remove this?

laws/shared/src/main/scala/algebra/laws/RingLaws.scala
```diff
@@ -201,6 +201,57 @@ trait RingLaws[A] extends GroupLaws[A] { self =>
...
+        A.equotmod(x, y)._2 ?== A.emod(x, y)
+      }
+    },
+    "euclidean function" -> forAll { (x: A, y: A) =>
+      pred(y) ==> {
+        val (_, r) = A.equotmod(x, y)
+        A.isZero(r) || (A.euclideanFunction(r) < A.euclideanFunction(y))
+      }
+    },
+    "submultiplicative function" -> forAll { (x: A, y: A) =>
```
johnynek, line 248: I didn't see this comment in trait. Could we add it so people get more intuition for the euclideanFunction?

core/src/main/scala/algebra/ring/DivisionRing.scala
```diff
@@ -0,0 +1,22 @@
+package algebra
+package ring
+
+import scala.{specialized => sp}
+
+trait DivisionRing[@sp(Byte, Short, Int, Long, Float, Double) A] extends Any with Ring[A] with MultiplicativeGroup[A] {
+  self =>
+
+  def fromDouble(a: Double): A = Field.defaultFromDouble[A](a)(self, self)
```
johnynek, line 9: Is there any law on this? Seems odd to me a division ring only has this extra method. We don't have any other division like operation?

core/src/main/scala/algebra/ring/EuclideanRing.scala
```diff
@@ -0,0 +1,59 @@
...
+ *   (for every x and non-zero y) x = yq + r, and r = 0 or f(r) < f(y).
+ *
+ * This generalizes the Euclidean division of integers, where f represents
+ * a measure of length (or absolute value), and the previous equation
+ * represents finding the quotient and remainder of x and y. So:
+ *
+ *   quot(x, y) = q
+ *   mod(x, y) = r
+ */
+trait EuclideanRing[@sp(Int, Long, Float, Double) A] extends Any with GCDRing[A] { self =>
```
johnynek, line 22: If we have specialized on these types can we also add the implementations?

johnynek (commented):
Looks cool!

core/src/main/scala/algebra/ring/EuclideanRing.scala
denisrosset: ... Float and Double are taken charge of by Field (euclidean division is trivial).

For Byte/Short/Int/Long, the instances are not lawful as overflow is undefined behaviour territory (which was not the case for just `CommutativeRing`, the laws pass).

We could:

1. Provide no instances
2. Provide instances, but not test them
3. Port the `forAllSafe` machinery from Spire that enables tests with abort-on-overflow

What do you think?

core/src/main/scala/algebra/ring/DivisionRing.scala
denisrosset: Field has it too. I should have `fromDouble` in DivisionRing and make `Field` inherit `DivisionRing`.

Basically, if you have a type `A` with a function `BigInt -> A` and division, you have `fromDouble`.

core/src/main/scala/algebra/ring/EuclideanRing.scala
denisrosset: Sure. Or I'll implement the extended GCD algorithm.

laws/shared/src/main/scala/algebra/laws/RingLaws.scala
denisrosset: Yes!

PR #247, opened 2021-05-14
Add Signed and TruncatedDivision typeclasses

denisrosset:
This follows the Spire implementations as well (we just put `Sign` into the `Signed` companion object).

We added the missing companion object functions.

### DIFF
--- a/core/src/main/scala/algebra/ring/Signed.scala
+++ b/core/src/main/scala/algebra/ring/Signed.scala
@@ -1,0 +1,141 @@
+package algebra
+package ring
+
+import scala.{specialized => sp}
+
+/**
+ * A trait that expresses the existence of signs and absolute values on linearly ordered additive commutative monoids
+ * (i.e. types with addition and a zero).
+ *
+ * The following laws holds:
+ *
+ * (1) if `a <= b` then `a + c <= b + c` (linear order),
+ * (2) `signum(x) = -1` if `x < 0`, `signum(x) = 1` if `x > 0`, `signum(x) = 0` otherwise,
+ *
+ * Negative elements only appear when the scalar is taken from a additive abelian group. Then:
+ *
+ * (3) `abs(x) = -x` if `x < 0`, or `x` otherwise,
+ *
+ * Laws (1) and (2) lead to the triange inequality:
+ *
+ * (4) `abs(a + b) <= abs(a) + abs(b)`
+ *
+ * Signed should never be extended in implementations, rather the [[Signed.forAdditiveCommutativeMonoid]] and
+ * [[Signed.forAdditiveCommutativeGroup subtraits]].
+ *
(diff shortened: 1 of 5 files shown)


core/src/main/scala/algebra/ring/Signed.scala
```diff
@@ -0,0 +1,141 @@
...
+  sealed abstract class Sign(val toInt: Int) {
+    def unary_- : Sign = this match {
+      case Positive => Negative
+      case Negative => Positive
+      case Zero     => Zero
+    }
+
+    def *(that: Sign): Sign = Sign(this.toInt * that.toInt)
+
+    def **(that: Int): Sign = this match {
```
johnynek, line 108: Can we add a comment that this is exponentiation?

laws/shared/src/main/scala/algebra/laws/OrderLaws.scala
```diff
@@ -112,6 +114,75 @@ trait OrderLaws[A] extends Laws {
     }
   )
 
+  def signed(implicit A: Signed[A]) = new OrderProperties(
+    name = "signed",
+    parent = Some(order),
+    "abs non-negative" -> forAll((x: A) => A.sign(A.abs(x)) != Signed.Negative),
```
johnynek, line 120: Do we require that abs(x) >= x?

Seems natural to me, but I don't see that is implied.

laws/shared/src/test/scala/algebra/laws/LawTests.scala
```diff
@@ -123,6 +123,8 @@ class LawTests extends munit.DisciplineSuite {
   checkAll("Long", RingLaws[Long].commutativeRing)
   checkAll("Long", LatticeLaws[Long].boundedDistributiveLattice)
 
+//  catsKernelStdOrderForBigInt
+  checkAll("BigInt", OrderLaws[BigInt].truncatedDivision)
```
johnynek, line 127: Can we add implementations for the specialized things?

johnynek (commented):
Very cool. Thanks for sending.

laws/shared/src/test/scala/algebra/laws/LawTests.scala
denisrosset: Same comment as for the intermediate ring structures-- the instances are only approximations due to the type limited range.

laws/shared/src/main/scala/algebra/laws/OrderLaws.scala
denisrosset: Sure!

core/src/main/scala/algebra/ring/Signed.scala
denisrosset: Sure!

PR #249, opened 2021-05-22
Suport scala 3

cquiroz:
Just bumping some dependencies


cquiroz:
Thanks, could we make a release with 3.0 support?

larsrh:
On its way to Central.

PR #252, opened 2021-11-05
replace traversable with iterable

i10416:
- [x] collection compat


armanbilge:
FYI algebra will no longer be developed here, but in the cats repo:
> The code in this repository was moved into Cats in cats#3918. All future development will occur there. Algebra will continue to release under the same artifact name and maintain binary compatibility with the 2.x series.

https://github.com/typelevel/algebra#current-status

@johnynek would you consider archiving this repo?

i10416:
Thanks for informing me of that!

